PR #21, opened 2023-03-23
Product milestone 1

brunobuddy:


ismaelguerrib (commented):
When I generate a resource there is no property name by default.

SebConejo:
@brunobuddy On the create-edit page, when we submit a form with a Date yield, an error says a date is expected when there is a date
Capture d’écran 2023-03-27 à 12 30 52

PR #102, opened 2023-12-30
credentails displayed on log while db is empty fixed

Tjs002:


brunobuddy:
Hello @Tjs002 

Thank you for your work ! CASE changed a lot: it is now Manifest and consists into a single YAML file where everything is. Unfortunately I have to close this PR for now. 

You can have a look at the new project: https://manifest.build/docs

PR #107, opened 2024-04-12
Feature/style manifest v0.1

SebConejo:
## Description
Set new style of admin panel
- Remove sidebar collapse feature
- Remove footer
- Set neutral design replacing the electric purple color by a black one.
- Update didebar design
- Refactor style code
- Remove header
- Delete images for a more neutral look


packages/core/manifest/backend.yml
```diff
@@ -0,0 +1,64 @@
+name: Blog about cats
```
brunobuddy: @SebConejo this file should be in the `manifest` folder to be taken in consideration. If it is unused you can delete it. 

The one in the manifest folder in in gitignore to prevent commiting it.

brunobuddy (changes requested):
A small request on a useless file. Everything else is flawless

brunobuddy (approved):

PR #154, opened 2024-09-23
(feat) sass refactor in  src/components/spinner.scss 

aditya-nv-06:
## Description
Refactoring the code based on the issue no #149 

## Impacted packages

Check the packages that require a new publication or release:

- [ ] @mnfst/admin

### DIFF
--- a/.github/workflows/label-pr.yml
+++ b/.github/workflows/label-pr.yml
@@ -1,0 +1,16 @@
+name: Add label to merged PRs
+
+on:
+  pull_request:
+    types: [closed]
+
+jobs:
+  add-label:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


SebConejo (changes requested):
@Aditya-NV-06 This PR breaks the code. Serving the admin panel generates a lot of errors. 

Since you're using @use instead of @import, the behavior is slightly different as @use creates a namespace for the imported file, and variables need to be referenced with that namespace. The error you're encountering likely stems from trying to directly use a variable as $grey-dark, which is not available globally unless imported or defined properly within the same scope.

As said by DM on Discord, I refuse this PR and I recommend you to try again once we pushed the next version. 

On this next version, it will be easier to contribute.

PR #158, opened 2024-09-24
(fix) code refactoring @import changed to @use

aditya-nv-06:
The pr regarding the issue #152 

Check the packages that require a new publication or release:

- [ ] @mnfst/manifest


SebConejo (changes requested):
This PR breaks the admin panel build. 

See my comment on the PR #154

PR #312, opened 2025-02-10
Update Manifest.ts

mcoquet:
login function always returned true. Added a check to return false if no token was found.

## Description
The login function had a hardcoded "return true" making it always succeed when attempting to login. 

## How can it be tested?
try to login with bad credentials and it should return false. 

- [ ] I created the related [changeset](https://github.com/changesets/changesets) for my changes with `npx changeset`
- [x] I created a PR for the [documentation](https://github.com/mnfst/docs) if necessary and attached the link to this PR


brunobuddy:
Good point @mcoquet ! Could you create the changeset doing `npx  changeset` and triggering a patch for the SDK package ? When it asks questions leave everything blank. Thank you

mcoquet:
> Good point @mcoquet ! Could you create the changeset doing `npx changeset` and triggering a patch for the SDK package ? When it asks questions leave everything blank. Thank you

hey! Lost track of this. Happy to do so, but having trouble running `npx changeset` I'm getting a `could not determine executable to run` message. Any thoughts?

brunobuddy:
@mcoquet i think you have to be at the repository root level to run the `npx changeset` command. It's a bit of a hassle at the beginning but it's amazingly simple to automatically ajust version numbers and connect to a CD to publish

brunobuddy (approved):

Issue #318, opened 2025-02-20
Admin should not be able to delete itself


brunobuddy:
From the admin panel you can remove yourself from the app (from settings > admin) which results in a strange behavior. That action should be limited

rohityadavv7:
@brunobuddy is this issue still open?

PR #321, opened 2025-02-22
feat: prevent admin users from deleting their own accounts

Trighap52:
## Description
This PR prevents an admin from deleting their own account while logged in. Previously, the admin could remove themselves from the app (from settings > admin) which resulted in a strange behavior. Now, a error is thrown if an admin tries to. I also updated the tests to correctly test the deletion (with a new mock and a new test case).
## Related Issues
* [Admin should not be able to delete itself #318 ](https://github.com/mnfst/manifest/issues/318)
## How can it be tested?
Try to delete yourself as an admin and an error shoul be thrown.
image

- [ ] I created a PR for the [documentation](https://github.com/mnfst/docs) if necessary and attached the link to this PR

### DIFF
--- a/packages/core/admin/src/app/modules/crud/views/list/list.component.html
+++ b/packages/core/admin/src/app/modules/crud/views/list/list.component.html
@@ -137,7 +137,10 @@
                 ></a>
                 <a
                   class="button is-light is-small is-circle"
-                  (click)="toggleDeleteModal(item)"
+                  [class.is-disabled]="entityManifest.slug === 'admins' && item['id'] === currentAdminId"
+                  [attr.aria-disabled]="entityManifest.slug === 'admins' && item['id'] === currentAdminId ? true : null"
+                  [attr.title]="entityManifest.slug === 'admins' && item['id'] === currentAdminId ? 'You cannot delete your own admin account' : null"
+                  (click)="entityManifest.slug === 'admins' && item['id'] === currentAdminId ? null : toggleDeleteModal(item)"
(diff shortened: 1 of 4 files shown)


brunobuddy:
Hello @Trighap52 thanks for your contribution.

There is a conflict that has to be resolved. Can you check it ? Also we would need to add tests for this behavior.

brunobuddy (changes requested):
See previous comment

SebConejo:
Hi @Trighap52 , thank you for your contribution. 

I just wanted to check in and see if you need any help or clarification to move this forward. Let me know if there’s anything We can do!

Looking forward to your updates whenever you get a chance. 🙂

brunobuddy (changes requested):
Hello @Trighap52 

Thank you for your changes, there is still an error in that code: it does not check that we are actually deleting ADMIN entity ! If I am logged as default admin with id 1 it will fail everytime I try to delete ANY item with ID 1.

Ensure that the "entity" param equals the admin slug (constants.ts) before.

 Also be the "user.user" sounds a bit bad, maybe you can call it "requestUser.user", not perfect but sounds a bit better

sahithvibudhi:
@Trighap52 are you still working on this? If not, I’d be happy to pick it up and help move it forward

brunobuddy:
@sahithvibudhi there is no news form @Trighap52 so you can take it from here if you want. Beware as the master branch changed quite a lot during that time, I suggest to start again from the last master version

PR #420, opened 2025-06-02
✨ UUIDs instead of incremental IDs

brunobuddy:
## Description

This PR replaces incremental integer IDs by UUIDs for better privacy and security. 

Instead of:
`/api/collections/users/1`

 you get:
`/api/collections/users/40a0bb98-ca2c-4c34-b5d0-57c91415c8f5`

:warning: Related doc PR to validate https://github.com/mnfst/docs/pull/32

## Related Issues

Discussion https://github.com/mnfst/manifest/discussions/359

## How can it be tested?

- Run the project and CRUD operations on entities
- Seed complex data structure with relations
- Visit admin panel
- Test with JS SDK with sandbox (see js-sdk package readme)

- [ ] I created a PR for the [documentation](https://github.com/mnfst/docs) if necessary and attached the link to this PR

### DIFF
--- a/.changeset/two-dogs-sniff.md
+++ b/.changeset/two-dogs-sniff.md
@@ -1,0 +1,6 @@
+---
+'manifest': minor
+'@mnfst/sdk': minor
+---
+
+Replaced incremental int IDs by UUIDs, thanks @jerryjappinen
--- a/package-lock.json
+++ b/package-lock.json
@@ -60,7 +60,9 @@
       }
     },
     "node_modules/@actions/http-client/node_modules/undici": {
-      "version": "5.28.5",
+      "version": "5.29.0",
+      "resolved": "https://registry.npmjs.org/undici/-/undici-5.29.0.tgz",
+      "integrity": "sha512-raqeBD6NQK4SkWhQzeYKd1KmIG6dllBOTt55Rmkt4HtI9mwdWtJljnrXjAFUBLTSN67HWrOIZ3EPF4kjUw80Bg==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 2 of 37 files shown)


SebConejo:
# ✅ UUID Test Checklist

#### 1. Golbal Tests
- [x] Entity relationships using UUIDs are tested ✅ 
- [x] Accessing `/api/collections/users/123` returns clean 400 error ✅ 
- [x] Logs and error messages do not expose sensitive UUIDs ℹ️ -> **See note 4 below**
- [x] Make unauthorized requests should provide a clear error ✅ 

#### 2. CRUD API Tests
- [x] Create a new entity (e.g., user) via POST ✅ 
- [x] Read an entity using UUID (`GET /api/collections/users/:uuid`) ✅ 
- [x] Update an entity via PATCH ✅ 
- [x] Delete an entity via DELETE ✅ 
- [x] API endpoints support only UUIDs and fail cleanly with legacy IDs ✅ 
- [x] Test creation and linking of related entities ✅ 

#### 3. Admin Panel Tests
- [x] url shows UUID instead of ID ✅ 
- [x] Can view/edit/delete a record ✅ 
- [x] Can create new entities in admin panel ✅ 

#### 4. JS SDK & Sandbox Tests
- [x] CRUD operations work via SDK (see `js-sdk` README) ✅ 
- [x] Relational data creation works via SDK ✅ 
- [x] Sandbox runs without errors in console ✅ 

#### 7. Developer Experience
- [x] Documentation is updated or PR linked ℹ️ **-> See note 3 below**
- ❌ API Documentation is updated or PR linked ❌ **-> See note 2 below**
- [x] README or changelog updated for SDK ✅

SebConejo:
### Notes
**1. Admin panel behavior**
  - [x] The fact that the items are sorted by UUID by default is a bit confused. Ater creating an itemn it's complicated to find it in the admin panel. This feature makes the experience in the administrator interface even more complex than it already was. It further underlines the need to be able to filter and sort entries. Some users have talked about this need on Discord or during calls, but there is no single issue or discussion about it to date that would suggest it’s an urgent need.

**2. CRUD API Tests**
  - [x] The provided doc API only accepts request by ID (Get, DELETE, PUT and PATCH) but we can't use it. So the result is an inconsistent deliverable. We should be able to query by UUID in the API doc we supply ourselves.
  
    Capture d’écran 2025-06-03 à 12 10 46
    
**3. Doc**
  - [x] In the doc, I saw this sentence below. It mentions an invremental ID. Is it still true?
  
    > This file will generate the Cat and Dog entity both with a name property. In Manifest by default all entities have an incremental id so you do not need to add it. They also have automatic createdAt and updatedAt columns that are not selected by default in the API requests.

**4. Security**
- [ ] ❗ Replace 403 with 404 to avoid exposing valid UUIDs

When accessing a resource by UUID (e.g. /collections/dogs/:uuid), the API returns a 403 Forbidden if the item exists but the user doesn’t have access.

This behavior confirms that the UUID is valid, which can be exploited to guess existing resources.

**Expected behavior:**
Return a 404 Not Found instead, even if the item exists but is forbidden. This prevents leaking information about existing UUIDs.

This change is especially important if the API handles sensitive resources.

SebConejo (changes requested):
Here’s a quick summary of my comments above:

- **Notes 1 and 4 aren’t critical at all**: we can discuss whether it’s worth opening an issue for them.

- **Note 2 seems important**, as the API documentation deliverable is currently broken.

- **Note 3**: we should update that sentence in the documentation to make it clearer.

brunobuddy:
@SebConejo fixed 1,2 and 3 points.

I created a card for point 4: #421 (it was already the case before switching to UUIDs)

SebConejo (approved):

PR #455, opened 2025-07-02
✨ Rate limiting

brunobuddy:
## Description

Added rate limiting 

Docs PR: https://github.com/mnfst/docs/pull/38 (I added a new "security" page under "self hosting". We will add other content there soon.

## How can it be tested?

- See PR in doc to see implementation
- Try to setup a ratelimit really low, ex: {limit: 1, ttl: 10000} and make several requests with Bruno. You should see 429 errors after passing the limit.

- [x] I created a PR for the [documentation](https://github.com/mnfst/docs) if necessary and attached the link to this PR


SebConejo (changes requested):
1. When Admin panel navigation hits the global rate limit, it blocks the UI-> I'm automatically logout and the message "You must be logged out".

It could be nice to have a correct flash message saying 'Too many requests.

SebConejo:
@brunobuddy I have another comment: Could it be a good thing to allow users to define a specific rate limiting for admins and authenticable users ? For example, in a company, 5 admins could navigate through the same app and make 10 requests each in 1 minute. That would be normal. But you don't want someone else making 50 requests in 1 minute. Or does it look too much, too tricky

brunobuddy:
@SebConejo I wanted to make a simple feature to start and add layers after. I think that each rule or endpoint could have it's own rate limiting, but I wanted to keep it simple for now.

PR #595, opened 2026-01-12
Fix duplicate code in create-manifest script

brunobuddy:
The prepublishOnly script now removes any existing starter folder before copying. This prevents the cp -r command from creating a nested starter/starter structure when the target directory already exists from a previous failed publish or local development.

Bumps version to 2.0.3.


brunobuddy:
We need to put back the publish.yml workflow for this npm package

PR #739, opened 2026-02-20
fix: center workspace content when sidebar is hidden

SebConejo:
## Summary
- Adds `.main-content--full { margin-left: 0 }` to `theme.css` so the workspace (agents list) page content is properly centered when no sidebar is present
- The rule existed in `layout.css` but was overridden by `theme.css` which re-declared `.main-content` with the sidebar margin later in the cascade

## Test plan
- [ ] Open the workspace/agents list page — content should be centered, not offset to the left
- [ ] Navigate to an agent detail page — sidebar should still appear and content should be offset correctly


SebConejo:
@brunobuddy It looks like the push to deploy is setup with a wrong instance 
image

PR #874, opened 2026-02-26
Add badge for related repositories in README

SebConejo:


brunobuddy (changes requested):
What is that badge about? The alternatives are not even consistent https://relatedrepos.com/gh/mnfst/manifest

plurch:
> What is that badge about? The alternatives are not even consistent https://relatedrepos.com/gh/mnfst/manifest

@brunobuddy @SebConejo 

I saw the feedback and looked into this. [Related Repos](https://relatedrepos.com/) computes relevance by looking at stars given to repos and computing co-occurrence with other repos. Results can be thought of as "users who expressed interest in this repository also expressed interest in these other repositories".

It looks like this repository was [previously a 1-file backend](https://github.com/mnfst/manifest/discussions/531). So many users who starred it in the past were also interested in backends and databases which is evident on the [results list for mnfst/manifest](https://relatedrepos.com/gh/mnfst/manifest).

There is no way to change star history when pivoting an existing repo. Over time results may become more related to the new functionality as more users give new stars. Let me know if that makes sense and if you have any other feedback. Thanks, Patrick

brunobuddy:
@plurch thanks for highlighting this. Indeed we probably have mixed results because of the pivot. Closing this PR for now, may be interested later though!

PR #963, opened 2026-03-04
feat: add API key source links in provider modals

guillaumegay13:
Add "Get API key" links below provider API key inputs in the routing and email provider modals, so users can quickly navigate to key creation pages.

  **Changes:**
  - Added provider-to-URL mapping in `ProviderSelectModal.tsx` and `EmailProviderModal.tsx`
  - Rendered conditional helper links below API key fields
  - Added styling and tests

### DIFF
--- a/.changeset/wise-oranges-double.md
+++ b/.changeset/wise-oranges-double.md
@@ -1,0 +1,5 @@
+---
+"manifest": patch
+---
+
+Add "Where to get API key" links below provider API key inputs in routing and email provider modals.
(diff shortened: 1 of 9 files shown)


brunobuddy (changes requested):
@guillaumegay13 there is too much noise in that PR: Agent.md, formatting many unrelated files.

You should only have very few files changes for that feature

guillaumegay13:
@brunobuddy I cleaned noise :
- removed AGENTS.md and the unrelated backend/plugin changes
- added a guard test so any routing provider that requires an API key must also define a `whereToGet` URL
- kept that constraint at the provider level rather than the model level, since the URL is provider-specific

Also added the whereToGet URL for Emails provider integration :
- extracted the email provider API-key metadata into a small shared config as well, so those links also live in one place and are testable

brunobuddy:
@guillaumegay13 there is still noise around that email provider modal (and conflicts). Make sure that we only change files related to the PR aim

guillaumegay13:
@brunobuddy I removed noises around the email providers modal and kept it minimal

brunobuddy (approved):

PR #964, opened 2026-03-04
fix: reconcile local API key on every boot to prevent 401

guillaumegay13:
In local mode, if the API key in `~/.openclaw/manifest/config.json` is regenerated
  after initial bootstrap (e.g. plugin reinstall), the new key is never stored in
  `agent_api_keys`. OtlpAuthGuard hashes the new Bearer token, finds no match, and
  returns 401 on proxy/OTLP endpoints.

  **Changes:**
  - Move key registration outside the early return so it runs on every boot
  - Use `upsert` instead of `insert` to handle existing rows when the key changes
  - Added test for key reconciliation when tenant already exists

### DIFF
--- a/.changeset/fix-local-api-key-reconciliation.md
+++ b/.changeset/fix-local-api-key-reconciliation.md
@@ -1,0 +1,5 @@
+---
+"manifest": patch
+---
+
+Fix local API key reconciliation on boot to prevent 401 errors and clean stale models.json entries on mode switch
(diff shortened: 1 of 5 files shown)


guillaumegay13:
Additional fix: clean stale models.json on mode switch

  When switching from cloud to local mode (or vice versa), OpenClaw's
  per-agent models.json retains the previous baseUrl (e.g.
  app.manifest.build). Since models.json takes priority over
  openclaw.json during model resolution, requests are sent to the
  wrong endpoint → 401.

  injectProviderConfig() now removes the manifest entry from each
  agent's models.json, forcing the model resolver to fall through to
  the runtime config which always has the correct URL.

PR #1048, opened 2026-03-09
fix: Anthropic adapter handling of tool calls in streaming responses

craigjmidwinter:
<h2>Problem</h2>
<p>When routing through Manifest's Anthropic adapter in streaming mode, tool calls are silently dropped. The stream correctly emits <code>finish_reason: tool_calls</code> but contains zero tool call data — no function name, no arguments, no ID. This causes OpenClaw (and any downstream consumer) to see a tool call signal with nothing to execute, resulting in a blocked/broken tool use turn.</p>
<p>The bug does not affect non-streaming requests (<code>fromAnthropicResponse</code> handles <code>tool_use</code> blocks correctly), which is why direct model calls appeared to work while routing through Manifest did not.</p>
<h2>Root Cause</h2>
<p><code>createAnthropicStreamTransformer</code> only handled two of Anthropic's stream event types:</p>

Event | Status
-- | --
message_start | ✅ Emits the opening role: assistant chunk
message_delta | ✅ Emits the finish reason chunk

---
## Summary by cubic
Fixes dropped tool calls in Anthropic streaming. The adapter now emits OpenAI-compatible `tool_calls` chunks and streams function args correctly so tool use works end-to-end.

- **Bug Fixes**
  - Emit `tool_calls` on `content_block_start` (`tool_use`) with correct ids, names, and OpenAI indices; track by content index and reset on `message_start`.
  - Stream `input_json_delta` as function arguments (including empty deltas) across multiple tool calls with proper per-call routing; map `stop_reason: tool_use` to `finish_reason: tool_calls`, include usage in the final chunk, and add a unit test asserting `content_block_start` emission.

- **Dependencies**
  - Pinned `esbuild` via `overrides` to `0.27.3` and refreshed the lockfile; removed unused public assets in `packages/openclaw-plugin` to reduce bundle size.

<sup>Written for commit 759aaa055248b679d90ad7737e668be787ef54f1. Summary will update on new commits.</sup>


brunobuddy:
@craigjmidwinter it seams that the CI failed

craigjmidwinter:
FYI @brunobuddy I think this should be fixed

PR #1185, opened 2026-03-17
fix(routing): prevent cross-auth credential fallback for explicit auth_type

### LINKED ISSUE #1183: routing: explicit api_key lookup can fall through to OpenAI subscription credentials
## Summary

When a request resolves to `auth_type = api_key`, provider credential resolution can still fall through to another auth record for the same provider/model. For OpenAI this can cause the normal API-key path to send a raw subscription token blob like `{"t":"..."}` as the bearer secret instead of a real API key.

This is separate from the analytics bug where failed fallback rows inherit the primary auth type. The problem here is the actual request path can use the wrong credential type.

## Evidence

Observed provider error on the primary request:

```json
{
  "error": {
    "message": "Incorrect API key provided: {\"t\":\"ey...",
    "type": "invalid_request_error"
  }
}
```

Image

Relevant code paths:

- `packages/backend/src/routing/resolve.service.ts`
- `packages/backend/src/routing/routing.service.ts`
- `packages/backend/src/routing/proxy/proxy.service.ts`
- `packages/backend/src/routing/openai-oauth.service.ts`

There is also an existing spec in `packages/backend/src/routing/__tests__/routing.service.spec.ts` that currently codifies this cross-auth fallback behavior.

## Expected behavior

When `preferredAuthType` is `api_key`, credential lookup should only return an `api_key` record. It should not silently fall through to `subscription`.

At minimum, the OpenAI API-key path should defensively reject blob-shaped JSON secrets instead of forwarding them as bearer tokens.

## Related

- #1173
- #1180

guillaumegay13:
## Summary
- make `getProviderApiKey(..., preferredAuthType)` strict when `preferredAuthType` is provided
- prevent explicit `api_key` lookups from silently decrypting `subscription` credentials (and vice versa)
- update routing service tests to cover strict behavior and add a backend changeset

## Why
Requests resolved as `auth_type=api_key` could still fall through to a subscription credential record. For OpenAI this can forward blob-shaped subscription secrets on the API-key path.

Fixes #1183

## Validation
- npm test --workspace=packages/backend -- routing.service.spec.ts --runInBand
- npm test --workspace=packages/backend -- proxy.service.spec.ts --runInBand

---
## Summary by cubic
Stop cross-auth credential fallback when an auth type is explicitly requested. Prevents OpenAI subscription secrets from being used on API-key routes and vice versa.

- **Bug Fixes**
  - Make `getProviderApiKey(..., preferredAuthType)` strict; no cross-auth fallback. Default stays prefer `api_key` when no type is provided.
  - Return null when the preferred auth record has no encrypted key (for both `api_key` and `subscription`).
  - Update routing tests and add a backend changeset; fix `manifest` release metadata. Aligns with Linear issue 1183.

<sup>Written for commit 2c7085e8cba2068003c574d4707603928f5eff35. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/tough-planets-shake.md
+++ b/.changeset/tough-planets-shake.md
@@ -1,0 +1,5 @@
+"manifest-backend": patch
+---
+
+Stop cross-auth credential fallback when an auth type is explicitly requested.
+Requests that resolve to `auth_type: api_key` now only use API key credentials (and likewise for `subscription`) instead of silently decrypting another auth record.
--- a/.changeset/tough-planets-shake.md
+++ b/.changeset/tough-planets-shake.md
@@ -1,4 +1,5 @@
-"manifest-backend": patch
+---
+"manifest": patch
 ---
 
 Stop cross-auth credential fallback when an auth type is explicitly requested.
--- a/packages/backend/src/routing/routing.service.spec.ts
+++ b/packages/backend/src/routing/routing.service.spec.ts
@@ -1885,7 +1885,7 @@ describe('RoutingService', () => {
       expect(result).toBe('sk-api-key');
     });
 
-    it('should fall back to subscription when preferred api_key has no encrypted key', async () => {
+    it('should return null when preferred api_key has no encrypted key', async () => {
       const { encrypt, getEncryptionSecret } = await import('../common/utils/crypto.util');
       const secret = getEncryptionSecret();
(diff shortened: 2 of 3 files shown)


guillaumegay13:
Fixed the malformed changeset frontmatter identified by cubic in commit 2c7085e8. @cubic-dev-ai please re-run review on the latest commit.

PR #1186, opened 2026-03-17
fix(proxy): normalize Mistral tool-call IDs in OpenAI-compatible payloads

### LINKED ISSUE #1182: proxy: Mistral fallback rejects OpenAI tool_call ids from reused transcripts
## Summary

When fallback reuses an OpenAI-style tool-call transcript against Mistral, the request can fail because Manifest forwards nested tool call ids unchanged. Mistral rejects OpenAI-style ids such as `call005AKQn2j5TEr4S6i3zNN59moT` and expects 9-character alphanumeric ids instead.

## Evidence

Observed final fallback error:

```json
{
  "object": "error",
  "message": "Tool call id was call005AKQn2j5TEr4S6i3zNN59moT but must be a-z, A-Z, 0-9, with a length of 9.",
  "type": "invalid_function_call",
  "code": "3280",
  "raw_status_code": 400
}
```

Relevant code paths:

- `packages/backend/src/routing/proxy/provider-endpoints.ts`
- `packages/backend/src/routing/proxy/provider-client.ts`
- `packages/backend/src/routing/proxy/proxy.service.ts`

## Expected behavior

One of these should happen consistently:

- normalize tool call ids when routing OpenAI-style transcripts to Mistral, including both `tool_calls[].id` and matching `tool` message `tool_call_id`
- detect tool-call transcript incompatibility up front and skip Mistral as a fallback target for those requests

## Minimal repro shape

```json
{
  "messages": [
    { "role": "assistant", "content": null, "tool_calls": [{ "id": "call005AKQn2j5TEr4S6i3zNN59moT", "type": "function", "function": { "name": "x", "arguments": "{}" } }] },
    { "role": "tool", "tool_call_id": "call005AKQn2j5TEr4S6i3zNN59moT", "content": "ok" }
  ]
}
```

guillaumegay13:
## Summary
- normalize tool-call ids to Mistral-compatible 9-char alphanumeric values when forwarding OpenAI-format chat payloads to Mistral
- preserve id linkage by rewriting both assistant `tool_calls[].id` and matching tool message `tool_call_id` using the same per-request mapping
- keep already-valid Mistral ids unchanged
- add regression tests covering normalization, linkage, and non-mutation

Closes #1182

## Why
Mistral rejects OpenAI-style tool call ids like `call005AKQn2j5TEr4S6i3zNN59moT`. Fallbacks reuse the original transcript across providers, so Mistral fallback attempts can fail deterministically on id format even when earlier fallback behavior is otherwise correct.

### Automated
```bash
npm test --workspace=packages/backend -- provider-client.spec.ts --runInBand
```
Expected: suite passes; includes tests asserting Mistral id normalization and preserved `tool_call_id` references.

### Manual repro
1. Configure a routing chain where a request with tool-call transcript can fallback to Mistral.
2. Send a conversation containing an assistant tool call id like `call005AKQn2j5TEr4S6i3zNN59moT` and a tool message referencing that id.
3. Verify the fallback request to Mistral succeeds schema validation (no `invalid_function_call` about id length/charset).
4. Verify non-Mistral OpenAI-compatible providers still receive unchanged tool call ids.

---
## Summary by cubic
Normalizes tool-call IDs to Mistral’s 9-character alphanumeric format when forwarding OpenAI-style chat payloads, preserving links between assistant tool calls and tool messages. Adds collision-safe ID generation so Mistral fallbacks don’t fail on ID format. Closes #1182.

- **Bug Fixes**
  - Rewrites assistant `tool_calls[].id` and matching `tool_call_id` via a per-request map; keeps valid 9-char IDs unchanged.
  - Pre-scans to reserve existing valid IDs and generates `tc` + base36 IDs (e.g., `tc0000001`), skipping collisions.
  - Applies only to `mistral`; passthrough for `openai` and `openrouter`.
  - Tests cover normalization, collision avoidance (including not clobbering later valid IDs), linkage preservation, and non-mutation.

<sup>Written for commit f1fac8cfa2d22a12e67e0bfc63dca47b2d56eaf9. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/few-cobras-wink.md
+++ b/.changeset/few-cobras-wink.md
@@ -1,5 +1,5 @@
 ---
-'manifest': patch
+"manifest": patch
 ---
 
 Normalize OpenAI-style tool call IDs when forwarding chat completions to Mistral so fallback requests remain compatible with Mistral's 9-character alphanumeric tool-call ID requirement.
--- a/.changeset/few-cobras-wink.md
+++ b/.changeset/few-cobras-wink.md
@@ -1,0 +1,5 @@
+---
+'manifest': patch
+---
+
+Normalize OpenAI-style tool call IDs when forwarding chat completions to Mistral so fallback requests remain compatible with Mistral's 9-character alphanumeric tool-call ID requirement.
--- a/packages/backend/src/routing/proxy/__tests__/provider-client.spec.ts
+++ b/packages/backend/src/routing/proxy/__tests__/provider-client.spec.ts
@@ -871,6 +871,64 @@ describe('ProviderClient', () => {
       expect(sentBody.messages[1].tool_call_id).toBe(validToolCallId);
     });
 
+    it('does not rewrite later valid Mistral tool call ids when generated ids would collide', async () => {
+      mockFetch.mockResolvedValue(new Response('{}', { status: 200 }));
+      const invalidToolCallId = 'call005AKQn2j5TEr4S6i3zNN59moT';
+      const validToolCallId = 'tc0000001';
+      const bodyWithPotentialCollision = {
+        messages: [
+          {
+            role: 'assistant',
+            content: null,
+            tool_calls: [
+              {
+                id: invalidToolCallId,
+                type: 'function',
+                function: { name: 'search', arguments: '{}' },
+              },
+            ],
+          },
+          {
+            role: 'tool',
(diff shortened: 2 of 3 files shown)


guillaumegay13:
Fixed the Mistral tool-call ID collision identified by cubic in commit b4ad6b9b. @cubic-dev-ai please re-run review on the latest commit.

PR #1187, opened 2026-03-17
proxy: add native OpenAI subscription Responses adapter support for tool calls

### LINKED ISSUE #1184: proxy: OpenAI subscription fallback fails on tool-call transcripts with null content
## Summary

When fallback lands on the OpenAI subscription / ChatGPT Codex path, tool-call transcripts can fail before the model runs because `assistant` messages with `content: null` are forwarded into the Responses-style payload unchanged. OpenAI then rejects the request with `input[n].content` invalid-type errors.

## Evidence

Observed fallback error:

```json
{
  "error": {
    "message": "Invalid type for 'input[3].content': expected one of an array of objects or string, but got null instead.",
    "type": "invalid_request_error",
    "param": "input[3].content",
    "code": "invalid_type"
  }
}
```

Capture d’écran 2026-03-17 à 21 11 54

Relevant code paths:

- `packages/backend/src/routing/proxy/chatgpt-adapter.ts`
- `packages/backend/src/routing/proxy/provider-client.ts`

The adapter already contains a comment noting that tool and function messages are not supported on this path.

## Expected behavior

One of these should happen consistently:

- translate tool-call transcripts into a valid Responses API request for the subscription path
- detect unsupported transcript shapes and skip OpenAI subscription fallbacks instead of attempting an invalid request

## Minimal repro shape

```json
{
  "messages": [
    { "role": "assistant", "content": null, "tool_calls": [{ "id": "call_123", "type": "function", "function": { "name": "x", "arguments": "{}" } }] },
    { "role": "tool", "tool_call_id": "call_123", "content": "ok" }
  ]
}
```

guillaumegay13:
## Summary
- route OpenAI `subscription` traffic through the ChatGPT Responses adapter while preserving the existing Chat Completions contract for Manifest clients
- translate Chat Completions tool definitions and tool history into native Responses API items (`tools`, assistant `tool_calls`, tool outputs), then translate Responses function calls back into Chat Completions, including streaming SSE events
- merge `system` and `developer` prompts into `instructions` and normalize null content so subscription primary and fallback requests no longer fail on `input[n].content` validation errors
- keep proxy fallback handling and metadata intact while allowing OpenAI subscription requests to proceed on the actual supported protocol
- add adapter and proxy regression coverage for null content, developer prompts, native tool calling, and streaming tool-call translation

Closes #1184.

## Why
OpenAI subscription auth does not speak `/v1/chat/completions`; it must use the ChatGPT Responses endpoint. The earlier guard-and-skip approach avoided some failures but did not actually restore tool use on the subscription path. This PR implements the protocol bridge so OpenAI subscription requests can use tools end-to-end without changing the external Chat Completions contract used by the rest of Manifest.

## Validation
- `npm test --workspace=packages/backend -- --runInBand routing/proxy/__tests__/chatgpt-adapter.spec.ts routing/proxy/__tests__/proxy.service.spec.ts`

### DIFF
--- a/.changeset/six-snakes-clean.md
+++ b/.changeset/six-snakes-clean.md
@@ -1,0 +1,5 @@
+---
+"manifest": patch
+---
+
+Add native OpenAI subscription Responses adapter support for tool calls, including tool transcript translation, streaming function-call events, and null or multipart content normalization.
--- a/packages/backend/src/routing/proxy/__tests__/chatgpt-adapter.spec.ts
+++ b/packages/backend/src/routing/proxy/__tests__/chatgpt-adapter.spec.ts
@@ -79,6 +79,17 @@ describe('ChatGPT Adapter', () => {
       expect(input[0].content[0].type).toBe('input_text');
     });
 
+    it('normalizes null content into an empty text part', () => {
+      const body = {
+        messages: [{ role: 'user', content: null }],
+      };
+      const result = toResponsesRequest(body, 'gpt-5');
+
+      expect(result.input).toEqual([
+        { role: 'user', content: [{ type: 'input_text', text: '' }] },
+      ]);
+    });
+
     it('remaps content to "output_text" for assistant messages', () => {
       const body = {
         messages: [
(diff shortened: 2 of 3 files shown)


guillaumegay13:
(!) After reproducing with the fix, it appears the issue isn't actually fixed yet

guillaumegay13:
Tested & validated manually. The fix is now cleaner than before - it integrates the OpenAI Responses API for OpenAI subscription and converts tools format between Chat Completion & Responses API formats so context is persevered when switching models during a conversation. FYI @brunobuddy 

docs here: https://developers.openai.com/api/docs/guides/migrate-to-responses

brunobuddy:
Finished in https://github.com/mnfst/manifest/pull/1386 @guillaumegay13 thanks!

Issue #1203, opened 2026-03-20
Alibaba provider with API Key fails "No endpoint configured for provider: qwen"


guillaumegay13:
<img width="1279" height="271" alt="Image" src="https://github.com/user-attachments/assets/db1549d9-ffd3-4d1d-9b57-b799f72a31c0" />
<img width="1161" height="343" alt="Image" src="https://github.com/user-attachments/assets/77e0d039-0e32-4677-9910-910273e58d58" />

guillaumegay13:
Whenever setting up its Alibaba account to generate an API Key, we need to select a region (China Beijing, Singapore, Hong Kong, US) and they seem to have a different URL.

PR #1206, opened 2026-03-20
fix: validate routing models against discovery

guillaumegay13:
---
## Summary by cubic
Enforces that routing selections only use models discovered for the agent and removes pricing/prefix-based fallbacks. Improves cleanup on provider removal while preserving native-only picks when an API key remains. Related to #1203.

- **Bug Fixes**
  - Validate `setOverride`/`setFallbacks` against discovery; reject missing models with `BadRequestException`.
  - `getEffectiveModel`: use the override only if it’s discovered; otherwise fall back to auto; remove pricing/provider-prefix heuristics.
  - On provider removal, clear overrides/fallbacks tied to the removed provider or its removed `cached_models`, and notify; keep native-only models if another active auth (e.g., API key) for the same provider remains.
  - Discovery: do not synthesize api-key models from pricing/OpenRouter when native discovery returns empty; subscription providers are still supplemented with known subscription models.
  - `setFallbacks`: de-duplicate inputs, exclude the current override, and return the cleaned list.

<sup>Written for commit 7d0f40e3e54c3429983c228b52905bce4188ccc2. Summary will update on new commits.</sup>

Related to #1203


guillaumegay13:
@brunobuddy merge conflicts resolved.

PR #1492, opened 2026-04-09
fix(manifest-model-router): use AUTO_MODEL in injectProviderConfig

### LINKED ISSUE #1464: manifest-model-router: injectProviderConfig uses minimal model object instead of AUTO_MODEL
### Plugin
`manifest-model-router` v6.0.2 (OpenClaw plugin)

### Description

The `injectProviderConfig` function in `manifest-model-router` writes a minimal model object `{id:"auto", name:"auto"}` to `openclaw.json` instead of using the already-defined `AUTO_MODEL` constant which includes the required `input`, `cost`, `contextWindow`, and `maxTokens` fields.

This causes OpenClaw to crash with:
```
Cannot read properties of undefined (reading 'input')
```

### Root Cause

In `dist/index.js`, the `injectProviderConfig` function hardcodes:
```js
let providerConfig = {baseUrl, api:"openai-completions", apiKey, models:[{id:"auto", name:"auto"}]}
```

But `AUTO_MODEL` is already correctly defined earlier in the same file:
```js
var AUTO_MODEL = {id:"auto", name:"Auto Router", reasoning:false, input:["text"], cost:{input:0, output:0, cacheRead:0, cacheWrite:0}, contextWindow:2e5, maxTokens:16384}
```

And is correctly used in `buildModelConfig` and the auth flow — but **not** in `injectProviderConfig`.

### Additional Issue

The `configSchema` in `openclaw.plugin.json` has `additionalProperties: false` but does not declare `apiKey` as an allowed property. The plugin code reads `config.apiKey`, but setting it via `openclaw config set plugins.entries.manifest-model-router.config.apiKey` fails config validation because the schema doesn't allow it.

### Fix

In `injectProviderConfig`, replace:
```js
models:[{id:"auto", name:"auto"}]
```
with:
```js
models:[AUTO_MODEL]
```

And add `apiKey` to the `configSchema.properties` in `openclaw.plugin.json`.

### Environment
- OpenClaw: 2026.3.31
- Plugin: manifest-model-router v6.0.2
- OS: Ubuntu Linux

thomas-norg-ai:
injectProviderConfig was writing a minimal model object {id:"auto", name:"auto"} missing the input, cost, contextWindow, and maxTokens fields. OpenClaw crashes with "Cannot read properties of undefined (reading 'input')" because the model catalog entry has no input field.

Export AUTO_MODEL from auth.ts and use it in provider-inject.ts so the injected config includes all required fields.

Fixes #1464

---
## Summary by cubic
Fixes an OpenClaw crash in the manifest model router by exporting and using the full `AUTO_MODEL` when injecting provider config, ensuring required fields are present. Also spreads `buildModelConfig(...)` during provider registration so `models` is a flat array, preventing a TypeError in provider resolution (fixes #1464).

<sup>Written for commit 120393491da635c8f564ca6e1a398588ff316330. Summary will update on new commits.</sup>

### DIFF
--- a/packages/openclaw-plugins/manifest-model-router/src/auth.ts
+++ b/packages/openclaw-plugins/manifest-model-router/src/auth.ts
@@ -1,7 +1,7 @@
 import { API_KEY_PREFIX, DEFAULTS } from './constants';
 import type { ProviderAuthContext, ProviderAuthResult } from './types';
 
-const AUTO_MODEL = {
+export const AUTO_MODEL = {
   id: 'auto',
   name: 'Auto Router',
   reasoning: false,
--- a/packages/openclaw-plugins/manifest-model-router/src/index.ts
+++ b/packages/openclaw-plugins/manifest-model-router/src/index.ts
@@ -5,7 +5,7 @@ import { registerCommand } from './command';
 import { verifyConnection } from './verify';
 import { injectProviderConfig, injectAuthProfile } from './provider-inject';
 import { stripOtlpSuffix } from './compat';
-import { runApiKeyAuth, buildModelConfig } from './auth';
+import { runApiKeyAuth, buildModelConfig, AUTO_MODEL } from './auth';
 import { ENV } from './constants';
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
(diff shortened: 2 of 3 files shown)


brunobuddy:
@thomas-norg-ai thanks for your contribution but we deprecated our openclaw plugin as it was creating many bugs for users. Docker is now the go-to way of self-hosting manifest

PR #1557, opened 2026-04-13
fix(docker): hardening phase 3 — closes #1551

### LINKED ISSUE #1551: Docker hardening follow-ups (Phase 3)
Follow-up polish items from the Docker release audit. The critical correctness fixes (routing status honesty, provider whitelist, tier override validation, pricing cache health UI, `.env.example` wiring, MANIFEST_TRUST_LAN docs, OpenClaw branding) are being shipped in the companion PR. This issue tracks the remaining items that didn't make that pass.

### Compose hardening leftovers (`docker/docker-compose.yml`)
- [ ] Add a size limit to the `/tmp` tmpfs: `tmpfs: - /tmp:size=64m`. A misbehaving upload or stream buffer today can grow unbounded into container memory.
- [ ] Raise `pids_limit` from 256 to 512. Node + Better Auth + proxy fanout can burst past 100 pids easily, and 256 is close enough to the ceiling to cause surprise EAGAIN failures.
- [ ] Replace the healthcheck `wget -qO-` with `node -e "fetch('http://127.0.0.1:3001/api/v1/health').then(r=>process.exit(r.ok?0:1)).catch(()=>process.exit(1))"`. BusyBox wget support for \`-qO-\` is shaky across Alpine variants; node is guaranteed to be present in the runtime image.

### HTTP status codes for non-chat error responses
- [ ] In \`packages/backend/src/routing/proxy/proxy-exception.filter.ts\`, detect the request intent (e.g. \`Accept: application/json\` vs. a chat client SSE/OpenAI client) and return real 4xx statuses (401/400) for auth/body errors when the caller clearly isn't a chat UI. Keep the current friendly HTTP 200 envelope for chat clients. CI pipelines and monitors are currently misled into thinking broken configs are healthy.

### Gate \`/api/v1/public/usage\` behind an opt-in env var
- [ ] \`packages/backend/src/public-stats/public-stats.controller.ts\` — only serve the public usage stats when \`MANIFEST_PUBLIC_STATS=true\`. Default off. Loopback binding in the Phase 1 PR already mitigates the default exposure, but anyone running Manifest behind a public reverse proxy today leaks aggregate message counts to unauthenticated callers.

### Verify + demote Better Auth \"User not found\" log level
- [ ] Reproduce by attempting login with an unknown email and grepping backend logs for \`User not found\`. Confirmed absent from our own source, so it's likely emitted upstream by Better Auth. If so, wrap their logger or file upstream. If it turns up in our own code, demote to WARN — it's an enumeration aid and floods alerting pipelines that watch for ERROR-level spikes.

### Parameterize \`og:url\` / \`og:image\` from \`BETTER_AUTH_URL\`
- [ ] \`packages/frontend/index.html\` currently hardcodes \`https://app.manifest.build\` for \`og:url\` and \`og:image\`. Inject them at boot time from \`BETTER_AUTH_URL\` (server-side template substitution) so self-hosters' shared links carry their own branding instead of Cloud Manifest's.

### CI smoke test for \`read_only: true\` compose boot
- [ ] Add \`.github/workflows/docker-smoke.yml\` that runs \`docker compose -f docker/docker-compose.yml up -d\`, waits for \`/api/v1/health\`, runs a login + agent-create e2e path via curl, and tears down. Guards against any future code that silently writes to disk (which would crash the read-only production container).

### Narrow the Dockerfile \`*.md\` cleanup inside node_modules
- [ ] \`docker/Dockerfile\` currently runs \`find . -path \"*/node_modules/*\" -name \"*.md\" -delete\`. Rare, but some packages read markdown at runtime (e.g. \`js-yaml\`'s schema docs). Scope the delete to package roots only (\`maxdepth\`-bounded find or explicit README targeting).

### Minor — filter diagnostic/error \`agent_messages\` rows out of the \"messages today\" counters
- _(already landed in the Phase 1/2 PR for workspace-card and per-agent analytics, not for \`/api/v1/messages\` itself)._ Consider adding a \`status\` filter parameter to \`/api/v1/messages\` so the UI can offer an \"errors only\" toggle without a schema change.

brunobuddy:
## Summary

Implements the seven follow-up items tracked in #1551 after the Phase 1+2 docker hardening pass shipped in #1552. Skips item 4 (Better Auth user-not-found ERROR log demotion) which needs runtime reproduction first.

### Compose hardening (`docker/docker-compose.yml` + `docker/Dockerfile`)
- `/tmp` tmpfs capped at `64m` so a misbehaving upload or stream buffer can't grow unbounded into container memory.
- `pids_limit` raised from 256 → 512. Node + Better Auth + the proxy's fanout routinely burst past 100 pids, and 256 was close enough to the ceiling to cause surprise EAGAIN failures.
- Healthcheck switched from BusyBox `wget -qO-` (compat-shaky across Alpine variants) to `node -e "fetch(...).then(r => process.exit(r.ok?0:1))..."` — guaranteed available in the runtime image. Updated in **both** the compose file and the in-image `HEALTHCHECK` directive.
- `node_modules` `*.md` cleanup narrowed to `README*` only so packages that read nested markdown at runtime (e.g. `js-yaml` schema docs) keep working.

### Public stats opt-in
`/api/v1/public/{usage,free-models,provider-tokens}` now require `MANIFEST_PUBLIC_STATS=true` to serve. Default off — returns 404 (not 403) so unauthenticated probes can't even tell the endpoints exist on a self-hosted instance. The flag is exposed via `app.publicStatsEnabled` in the existing `appConfig` so it's testable through `ConfigService`.

### Honest HTTP statuses for tool callers (`proxy-exception.filter.ts` + `proxy.controller.ts`)
- New `isChatRenderingClient(req)` helper detects chat clients via `body.stream === true` OR `Accept: text/event-stream`.
- Non-chat clients (curl/CI/monitors/non-streaming SDKs) now get **real** `401`/`400`/`500` HTTP statuses with a structured `{ error: { message, type, code? } }` envelope instead of a friendly HTTP-200 stub. CI pipelines watching status codes will finally detect broken setups.
- Chat clients (anything sending `stream: true` or `Accept: text/event-stream`) continue to receive the friendly HTTP-200 envelope so the rendered assistant message keeps working.
- Both the `@UseFilters(ProxyExceptionFilter)` decorator AND the `proxy.controller`'s inline `try/catch` block were updated, since the catch swallows BadRequestException before it reaches the filter.

### og: tag personalization
- New `rewriteOgTags` helper replaces the SPA `index.html`'s hardcoded `https://app.manifest.build` with `BETTER_AUTH_URL` at boot time.
- `SpaFallbackFilter` performs the substitution once when caching `index.html`, then serves the rewritten content for every SPA fallback request.
- When `BETTER_AUTH_URL` is empty or matches the default, the rewrite is a no-op (Cloud Manifest behavior unchanged).

### `/api/v1/messages` status filter
New `status` query parameter accepting `ok | error | rate_limited | fallback_error | errors`, where `errors` expands to the union of the three error variants. Lets the dashboard build an "errors only" toggle without a schema change. The count cache key is extended so different statuses don't share counts.

### CI smoke test (`.github/workflows/docker-smoke.yml`)
- Builds the production image as `manifestdotbuild/manifest:latest`.
- Boots `docker compose up -d` with `read_only: true`.
- Polls `/api/v1/health` for up to 2 minutes.
- Asserts the SPA root loads and `/public/usage` is gated to 404 by default.
- Tears down on success or failure.
- Triggers on PRs touching `docker/`, `packages/`, or `package*.json` so any future code that silently writes to disk crashes CI before it crashes users.

## Test plan

- [x] `npm test --workspace=packages/backend` — **3556/3556 passing** (added 17 new tests across DTO/filter/controller/spec)
- [x] `npm run test:e2e --workspace=packages/backend` — **113/113 passing**
- [x] `npm test --workspace=packages/frontend` — **2110/2110 passing**
- [x] TypeScript clean on backend + frontend
- [x] Coverage on modified files: `og-rewrite.ts` 100%, `spa-fallback.filter.ts` 100%, `app.config.ts` 100%, `public-stats.controller.ts` 100% lines, `proxy-exception.filter.ts` 100%, `proxy.controller.ts` 100% lines, `messages-query.dto.ts` covered via expanded `messages-query.dto.spec.ts`
- [x] `docker compose config` resolves cleanly with new healthcheck array, `tmpfs: /tmp:size=64m`, `pids_limit: 512`
- [x] Manual smoke against `node dist/main.js`:
  - `/api/v1/public/{usage,free-models,provider-tokens}` → **404** by default; → **200** with `MANIFEST_PUBLIC_STATS=true`
  - `POST /v1/chat/completions` with `Authorization: Bearer mnfst_invalid` → **401** with `{error.type: 'auth_error'}`
  - Same with `Accept: text/event-stream` → **200** SSE friendly response
  - `curl /some-spa-route` with `BETTER_AUTH_URL=https://manifest.example.com` → og: tags rewritten to `manifest.example.com`; with `BETTER_AUTH_URL` unset → tags stay as `app.manifest.build`

## Notes

- `packages/manifest` changeset is included (`patch` bump). Closes #1551.
- The Better Auth "User not found" ERROR log demotion (item 4 from the issue) is intentionally deferred — it needs runtime reproduction to confirm whether the log originates upstream or in our own code.

---
## Summary by cubic
Phase 3 Docker hardening with automatic DB migrations and honest HTTP errors for tools. Improves self‑hosting safety and keeps public stats private by default. Closes #1551.

- **New Features**
  - Docker/Compose: cap `/tmp` tmpfs at 64MB, raise `pids_limit` to 512, switch healthchecks to `node -e fetch`, and only delete `README*` markdown in `node_modules`.
  - CI: new Docker Smoke workflow builds the image, boots with `read_only: true`, waits for `/api/v1/health`, verifies the SPA loads, and confirms `/public/usage` is 404 by default.
  - Public stats: `/api/v1/public/{usage,free-models,provider-tokens,free-providers}` gated behind `MANIFEST_PUBLIC_STATS=true` (default off → 404). Exposed via `app.publicStatsEnabled`.
  - HTTP errors: non‑chat callers (no `stream: true` and no `Accept: text/event-stream`) now get real `401`/`400`/`500` with `{ error: { message, type, code? } }`. Chat clients keep the friendly HTTP‑200 envelope.
  - SPA meta: rewrite `og:url`/`og:image` in `index.html` at boot using `BETTER_AUTH_URL` (no‑op when empty/default) so self‑hosted link previews use the correct URL.
  - Analytics: `/api/v1/messages` adds `status=ok|error|rate_limited|fallback_error|errors`; counts segment by status.
  - Database: always run TypeORM migrations on boot; dropped `AUTO_MIGRATE` (removed from compose/docs) so fresh installs start with the full schema.

<sup>Written for commit fbca5dc6e166567b989048b6f7296de88a91d4de. Summary will update on new commits.</sup>


brunobuddy:
Added one follow-up commit (8df6e3de): **always run TypeORM migrations on boot, drop the `AUTO_MIGRATE` gate**.

### Why
Another pair of eyes flagged that `manifestdotbuild/manifest:5.45.1` (the previously-published Docker tag) gates `migrationsRun` on `NODE_ENV === 'development' | 'test'` only, so a fresh production install only gets Better Auth's 4 tables — the dashboard 500s on `user_providers`/`notification_rules`. The fix landed in `fc7890fd` (local-mode removal) which added an `AUTO_MIGRATE === 'true'` short-circuit, and that's what the current `latest` tag (5.46.2) ships. So the bug is *already gone* — but pinning behavior to an env var that compose has to set is a footgun: a future env-var typo silently re-introduces it.

### What
- `database.module.ts`: `migrationsRun: true` unconditionally. Comment explains why (synchronize is permanently off, committed migrations are the only schema source, no production scenario where booting with pending migrations is correct).
- Removed `autoMigrate` field from `app.config.ts` and its three spec tests (nothing in production code reads it — the decision lives in `database.module.ts` which read `process.env` directly).
- Removed `AUTO_MIGRATE=true` from `docker-compose.yml` and the three `docker run` examples in `DOCKER_README.md` (now redundant).
- Removed the `AUTO_MIGRATE` comment from `packages/backend/.env.example`.

### CI note
Workflow runs for this commit fired via manual `workflow_dispatch` (the `pull_request:synchronize` trigger appears to have hiccupped — check #24359293197 / #24359294366 / #24359295493). All three pass:
- **CI** (lint + backend-postgres + frontend + shared) — success
- **Docker** (validate amd64 + arm64) — success
- **Docker Smoke** (read-only compose boot) — success

Local: 3555 backend unit + 113 e2e + 2110 frontend tests passing, typecheck clean.

SebConejo:
@brunobuddy What about this PR ? Can I review and merge it?

PR #1607, opened 2026-04-18
feat: make PROVIDER_TIMEOUT_MS configurable via environment variable

### LINKED ISSUE #1583: Feature: Make PROVIDER_TIMEOUT_MS configurable via environment variable
### Problem

`PROVIDER_TIMEOUT_MS` is hardcoded to `180_000` (180s) in `packages/backend/dist/routing/proxy/provider-client.js`. This value is not configurable via environment variable, database setting, or config file.

This creates two problems for self-hosted users running Manifest with providers that have unreliable response times (e.g. Ollama Cloud):

**1. Fallback chains become ineffective**

With 180s per attempt, a tier with 5 fallback models needs up to 180s × 6 = 18 minutes to exhaust the chain. In practice, the upstream client (e.g. OpenClaw gateway) times out long before Manifest reaches a working fallback.

**2. Timeout race with upstream clients**

OpenClaw's default `timeoutSeconds` is 180 — identical to Manifest's internal timeout. When both fire simultaneously, the client closes the connection first. Manifest then sees `signal.aborted = true` from the client disconnect, re-throws instead of falling back, and the fallback chain never runs. The logs show "Proxy error: The operation was aborted due to timeout" but no "Provider transport failure" entries, confirming the fallback path is bypassed.

### Root cause analysis

We traced the flow through the proxy code:

- `provider-client.js`: `fetch()` uses `AbortSignal.timeout(PROVIDER_TIMEOUT_MS)` — this is a total timeout (from request start), not an idle timeout
- `proxy-fallback.service.js`: `tryForwardToProvider()` catches the error
- `proxy-transport.js`: `isTransportError()` checks for `/timeout/i` pattern → creates synthetic 504
- `proxy.service.js`: `shouldTriggerFallback(504)` → true → fallback chain runs

This flow works correctly when Manifest's timeout fires before the client disconnects. But with matching timeouts (180s/180s), it's a race condition that Manifest usually loses.

Additional finding: mid-stream hangs (provider returns 200 OK, starts streaming, then stops sending chunks) are architecturally not fallback-capable. Once `headersSent = true`, the controller calls `res.end()` on timeout — no fallback path exists.

### Proposed solution

Add an environment variable `PROVIDER_TIMEOUT_MS` (or `PROVIDER_REQUEST_TIMEOUT`) that overrides the hardcoded value:

```javascript
const PROVIDER_TIMEOUT_MS = parseInt(process.env.PROVIDER_TIMEOUT_MS, 10) || 180_000;
```

This allows self-hosted users to set a lower timeout so the fallback chain can actually run within the upstream client's timeout window:

```yaml
# docker-compose.yml
environment:
  PROVIDER_TIMEOUT_MS: 45000
```

With 45s per attempt: primary (45s) + fallback 1 (45s) + fallback 2 (45s) = 135s total — well within a typical 300s upstream timeout, and the agent gets a response instead of an error.

### Environment

- Manifest: Docker (self-hosted, local mode)
- Upstream: OpenClaw 2026.4.14
- Affected providers: Ollama Cloud (glm-5.1:cloud, qwen3.5:cloud) — frequent silent hangs with no HTTP error, just open connections producing no data
- Fallback targets: Anthropic, OpenRouter, OpenAI — all functional but never reached due to timeout race

brian-c11:
## Summary

Fixes #1583

- Replaces the hardcoded `180_000` ms timeout in `provider-client.ts` with a value read from `process.env.PROVIDER_TIMEOUT_MS`, falling back to `180_000` if unset or invalid.

## Problem

With matching 180s timeouts between Manifest and upstream clients (e.g. OpenClaw), both fire simultaneously. The client disconnects first, causing Manifest to see `signal.aborted = true` from the client disconnect and re-throw instead of triggering the fallback chain.

## Solution

```typescript
const PROVIDER_TIMEOUT_MS = parseInt(process.env.PROVIDER_TIMEOUT_MS ?? '', 10) || 180_000;
```

Self-hosted users can now set a lower timeout in `docker-compose.yml`:

```yaml
environment:
  PROVIDER_TIMEOUT_MS: 45000
```

With 45s per attempt: primary (45s) + fallback 1 (45s) + fallback 2 (45s) = 135s — well within a typical 300s upstream timeout, allowing the fallback chain to complete.

## Test plan

- [ ] No `PROVIDER_TIMEOUT_MS` env var set → timeout defaults to `180_000`
- [ ] `PROVIDER_TIMEOUT_MS=45000` → timeout uses `45000`
- [ ] `PROVIDER_TIMEOUT_MS=abc` (invalid) → timeout falls back to `180_000`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Make the provider request timeout configurable via `PROVIDER_TIMEOUT_MS` instead of a fixed 180s. The value must be a finite positive integer; otherwise we default to 180000 ms to prevent proxy crashes and allow fallbacks, with `.env.example` guidance and tests covering default, valid, and invalid values.

<sup>Written for commit fbd445eabe6aebdb9e8bb77253d21a417783a19b. Summary will update on new commits. <a href="https://cubic.dev/pr/mnfst/manifest/pull/1607?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/packages/backend/.env.example
+++ b/packages/backend/.env.example
@@ -22,6 +22,11 @@ BETTER_AUTH_URL=http://localhost:3001
 # ── API Key (optional) ──────────────────────────────
 # API_KEY=                   # Secret for programmatic access via X-API-Key header
 
+# ── LLM Proxy ────────────────────────────────────────
+# Per-attempt timeout (ms) for upstream provider requests in the LLM proxy.
+# Set strictly below your client's timeout so the fallback chain has room to run.
+# PROVIDER_TIMEOUT_MS=180000
+
 # ── Email provider (optional, unified) ──────────────
 # Used for BOTH Better Auth transactional emails (login verification,
 # password reset) and threshold alert notifications when no per-user
--- a/packages/backend/src/routing/proxy/__tests__/provider-client.spec.ts
+++ b/packages/backend/src/routing/proxy/__tests__/provider-client.spec.ts
@@ -1739,4 +1739,59 @@ describe('ProviderClient', () => {
       ).rejects.toThrow('key=***');
     });
   });
+
+  describe('PROVIDER_TIMEOUT_MS env override', () => {
+    const originalEnv = process.env.PROVIDER_TIMEOUT_MS;
+
+    afterEach(() => {
+      if (originalEnv === undefined) delete process.env.PROVIDER_TIMEOUT_MS;
+      else process.env.PROVIDER_TIMEOUT_MS = originalEnv;
+    });
+
+    async function captureTimeoutMs(envValue: string | undefined): Promise<number> {
+      if (envValue === undefined) delete process.env.PROVIDER_TIMEOUT_MS;
+      else process.env.PROVIDER_TIMEOUT_MS = envValue;
+
+      const timeoutSpy = jest.spyOn(AbortSignal, 'timeout');
+      timeoutSpy.mockClear();
+
+      let observed = -1;
+      await jest.isolateModulesAsync(async () => {
+        const { ProviderClient: FreshClient } = await import('../provider-client');
+        const fresh = new FreshClient();
+        mockFetch.mockResolvedValueOnce(new Response('{}', { status: 200 }));
+        await fresh.forward({
+          provider: 'openai',
+          apiKey: 'sk-test',
+          model: 'gpt-4',
+          body,
+          stream: false,
+        });
+        observed = timeoutSpy.mock.calls[0]?.[0] ?? -1;
+      });
+
+      timeoutSpy.mockRestore();
+      return observed;
+    }
+
+    it('defaults to 180000 ms when env var is unset', async () => {
+      expect(await captureTimeoutMs(undefined)).toBe(180_000);
+    });
+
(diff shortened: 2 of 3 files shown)


brunobuddy:
Pushed two small additions on top of your commit (maintainer push, your authorship preserved):

1. **Hardened parsing** — addressed cubic's P1. `parseInt('-1', 10) = -1` is truthy, so the previous one-liner let negative values through to `AbortSignal.timeout(-1)`, which throws `TypeError` and crashes every proxy request. Now requires a finite positive integer; falls back to `180_000` otherwise.
2. **Documentation** — added `PROVIDER_TIMEOUT_MS` to `packages/backend/.env.example` with guidance to set it strictly below the upstream client's timeout so the fallback chain has room to run.
3. **Tests** — five unit tests covering default / valid override / non-numeric / negative / zero, in `provider-client.spec.ts`.

All 3646 backend tests pass; tsc clean. Docs PR for `mnfst/docs` going up next so self-hosted operators can find this knob.

brunobuddy:
Thanks @brian-c11 ! I added few tests to it and merged it

PR #1614, opened 2026-04-19
fix(proxy): preserve full model id for custom provider endpoints

xyz0z0:
Custom gateways configure upstream model names like z-ai/step-3.5-flash. stripModelPrefix was stripping the first path segment for all endpoints except openrouter, which broke custom HTTP backends that require the full string. Treat custom the same as openrouter for model passthrough.

Adds a regression test in provider-client.spec.ts.

Made-with: Cursor

---
## Summary by cubic
Preserves full vendor/path model IDs for custom provider endpoints to avoid breaking custom gateways (e.g., "z-ai/step-3.5-flash"). Aligns custom endpoint behavior with OpenRouter by skipping prefix stripping.

- **Bug Fixes**
  - Treat `custom` endpoints like `openrouter` in model handling to bypass prefix stripping.
  - Add regression test to ensure the full model ID is sent for custom endpoints.

<sup>Written for commit 0f4568c9e566dcf62a0f3577ee78b584c6b7dcc5. Summary will update on new commits.</sup>

### DIFF
--- a/packages/backend/src/routing/proxy/__tests__/provider-client.spec.ts
+++ b/packages/backend/src/routing/proxy/__tests__/provider-client.spec.ts
@@ -1614,6 +1614,32 @@ describe('ProviderClient', () => {
       expect(sentBody.stream).toBe(false);
     });
 
+    it('preserves vendor/path model ids for custom endpoints (no prefix stripping)', async () => {
+      mockFetch.mockResolvedValue(new Response('{}', { status: 200 }));
+
+      const customEndpoint = {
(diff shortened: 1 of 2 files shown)


guillaumegay13:
Closing this as superseded. The underlying custom-provider slash-model bug is real, but current main already fixes it in d5b23dc03 (fix(proxy): forward multi-segment custom-provider model ids unchanged), with broader regression coverage and a changeset. Thanks for reporting/fixing the right issue here.

PR #1652, opened 2026-04-21
feat(providers): add NanoGPT built-in provider

Daltonganger:
## Summary
- add NanoGPT as a built-in OpenAI-compatible provider across the shared registry, backend model discovery, proxy routing, and frontend provider UI
- add the NanoGPT API key URL and update provider-count expectations and focused provider endpoint coverage in backend and frontend tests
- verify the change with the backend and frontend test suites

---
## Summary by cubic
Adds `NanoGPT` as a built-in OpenAI-compatible provider so users can add an API key and use `/v1` endpoints via our proxy and UI. Also fixes shared model inference to route `nano-gpt/...` models to the right provider.

- **New Features**
  - Shared: added `nano-gpt` to the provider registry with alias `nanogpt`.
  - Backend: model discovery at https://nano-gpt.com/api/v1/models; proxy base https://nano-gpt.com/api with OpenAI headers/paths.
  - Frontend: provider UI entry, logo, API key link https://nano-gpt.com/settings/api-keys, and updated provider ordering.
  - Tests: provider counts updated to 16; added endpoint/header checks, alias resolution parity, and icon test updated to accept ``.

- **Bug Fixes**
  - Shared: infer `nano-gpt/` model prefixes to `nano-gpt`.

<sup>Written for commit 854a0477aa65a2da870a52112852e3e25e882302. Summary will update on new commits.</sup>

PR #1659, opened 2026-04-22
feat(providers): add NanoGPT subscription mode

Daltonganger:
## Summary
- add NanoGPT subscription-mode support using the dedicated subscription discovery and proxy endpoints
- show NanoGPT in the Subscriptions tab while keeping the existing API-key/paygo flow unchanged
- cover the new subscription routing and provider metadata with backend, shared, and frontend tests

## Notes
- This PR is stacked on top of #1652 because GitHub cannot target a fork-only branch as the base for an upstream PR.
- Until #1652 merges, GitHub will show the NanoGPT provider commits here too.
- After #1652 merges, this PR reduces to the subscription-specific follow-up.

---
## Summary by cubic
Adds NanoGPT as a built-in provider with a subscription mode using dedicated discovery and proxy endpoints. NanoGPT now shows in the Subscriptions tab, while the existing API‑key flow stays the same.

- **New Features**
  - Registered `nano-gpt` (alias: `nanogpt`) and updated provider resolution and counts.
  - Added OpenAI‑compatible proxy endpoints for `nano-gpt` and `nano-gpt-subscription`, with subscription routing from `nano-gpt` to `nano-gpt-subscription`.
  - Implemented model discovery configs for both modes and route to the subscription endpoint when `authType=subscription`.
  - Frontend updates: NanoGPT icon, provider UI with token-based subscription, API key URL mapping, and listing in both API Keys and Subscriptions tabs.
  - Shared subscription config for NanoGPT (capabilities, labels). Added tests across backend, shared, and frontend for endpoints, routing, inference, and UI.

<sup>Written for commit c6f9e22afa4b7f78141a238f028b91e379ebecc1. Summary will update on new commits.</sup>


brunobuddy:
@Daltonganger zero users added NanoGPT as their custom provider. 

If it's a compatible OpenAI API we would have seen many users adding it as custom providers. That is not the case,. Moreover we have some providers that have hundreds of custom implementations and not available on our curated list yet (kilo, groq, nvidia).

PR #1692, opened 2026-04-23
feat(benchmark): side-by-side model comparison

brunobuddy:
## Summary

Adds a new **Benchmark** page under the existing MANAGE sidebar section. Users can send a single prompt to 2-6 models in parallel and compare **cost, output tokens, and latency** side by side. Responses render as sanitized markdown (with syntax-highlighted code blocks via the existing \`highlight.js\`).

- **Columns**: horizontal scroll, one per model, each with live skeleton → response → metrics footer → collapsible response headers. Per-column retry and an "× remove" affordance.
- **Summary table**: Provider (with logo) · Model · Cost · Output tokens · Duration. Cheapest/fastest cells highlighted; non-winners show \`+X%\` deltas vs the winner.
- **Default state**: prefills two models from two different providers (when available) so an empty-page visitor can immediately see the concept.
- **History drawer**: every run is persisted (\`benchmark_runs\` + \`benchmark_columns\`, capped at 50 per user+agent, oldest pruned). Click a past run → replays into the current view.
- **Request headers** (power-user): subtle \`{ }\` icon left of **Send** opens a popover with a key/value row editor. Count badge when headers are set. Persisted to \`localStorage\`. Auth, transport-layer, and \`x-manifest-*\` headers are dropped both client-side (with an inline warning) and server-side.
- **Backend**: new \`BenchmarkModule\` with \`POST /api/v1/benchmark/run\`, \`GET /api/v1/benchmark/runs\`, \`GET /api/v1/benchmark/runs/:id\`. Reuses \`ProviderClient.forward\` and \`ProxyMessageRecorder\` so benchmark calls also show up in the regular Messages log with \`routing_reason='benchmark'\`. Response headers are whitelisted; request headers are sanitized. One new migration: \`1777100000000-AddBenchmarkHistory\`.
- **Icons**: the app's self-hosted boxicons font only ships 5 glyphs (check-circle, florist, info-circle, message-bubble-detail, x-circle); everything else introduced here is inline SVG in \`components/benchmark/icons.tsx\` matching the \`ProviderIcon.tsx\` pattern — no CDN, CSP-safe.

## Test plan

- [x] Backend unit tests (219 suites / 4,118 tests) — new: \`benchmark.service.spec.ts\`, \`benchmark-history.service.spec.ts\`, \`request-header-sanitizer.spec.ts\`
- [x] Backend e2e (17 suites / 132 tests, serial) — new: \`benchmark-flow.e2e-spec.ts\` covering happy path, cross-user 404, malformed run id 400, list + replay
- [x] Frontend tests (119 suites / 2,332 tests) — new: \`benchmark-store\`, \`BenchmarkSummaryTable\`, \`BenchmarkHistoryDrawer\`, \`MarkdownContent\` (including XSS strip), \`RequestHeadersPopover\`
- [x] \`tsc --noEmit\` clean in backend, frontend, and shared
- [x] \`eslint\` clean on every new file
- [x] Smoke-tested locally at http://localhost:54212: golden path, error path with inline retry, history drawer replay, request-headers popover with localStorage persistence, markdown rendering of a real provider response
- [ ] Reviewer: verify the new sidebar link shows up between Routing and Limits and that Benchmark runs appear in the Messages log with \`routing_reason='benchmark'\`

---
## Summary by cubic
Adds a Benchmark page to compare 2–6 models side by side and a replay flow that pins a recorded “Original” call. Also adds opt‑in per‑agent message recording with a redesigned drawer and promotes Benchmark to a first‑class message tier with a new filter and badge in Messages.

- **New Features**
  - Benchmark page: one prompt → parallel columns per model with retry/remove, collapsible headers, and horizontal scroll. Summary table highlights cheapest/fastest; non‑winners show +X% deltas. Default view preselects two models and prefers native‑branded picks from connected providers over aggregator proxies.
  - Replay a recorded query: pick any recorded call as a read‑only “Original” column, then Re‑run across chosen models. Backend forwards the recorded request body verbatim (minus `stream`/`stream_options`/`model`) so every column uses the same prompt/tools/params. Summary shows signed deltas versus Original; Original is excluded from winner highlights. Adds an Optimize button and `?optimize=<id>` deep link from the recording drawer that pins the Original and pre‑fills request headers.
  - History and headers: history drawer (up to 50 runs per user+agent) with replay. Request headers popover saved to `localStorage`; `authorization`, cookies, transport‑layer, and `x-manifest-*` are blocked client‑ and server‑side.
  - Rendering and API: responses render as sanitized markdown with code highlighting using `marked`, `dompurify`, and `highlight.js`. Backend `BenchmarkModule` exposes `POST /api/v1/benchmark/run`, `GET /api/v1/benchmark/runs`, `GET /api/v1/benchmark/runs/:id`; reuses `ProviderClient.forward`; whitelists safe response headers; benchmark calls write `routing_tier='benchmark'`. DECIMAL columns now deserialize as numbers for correct cost/threshold math.
  - Message recording (opt‑in): per‑agent toggle in Settings captures request body, response body, and response headers for proxy calls (2 MB cap) into a separate `message_recordings` table; excluded from anonymous telemetry. Messages log adds a "Recorded only" filter, a record‑dot row icon, and a right‑side drawer with tabs (Parameters, Conversation, Tool calls, Reply, Usage, Headers, Raw), plus delete support and an Optimize action.
  - Messages list: adds a Tier filter with a Benchmark badge and an Include‑benchmarks option (off by default).

- **Migration**
  - Run DB migrations: `1777110000000-AddAgentRecordMessages` (renamed; uses IF NOT EXISTS), `1777200000000-AddMessageRecordings`, `1777300000000-AddBenchmarkHistory`, `1777400000000-BackfillBenchmarkTier`.

<sup>Written for commit 15a1dbb95a17afff13eb9b00d1cbd0be53318683. Summary will update on new commits.</sup>


guillaumegay13:
@brunobuddy we can close it as you shipped the Playground already ?

PR #1693, opened 2026-04-23
feat(messages): opt-in per-agent message recording

brunobuddy:
## Summary

Adds an opt-in, per-agent **Record messages** toggle (default off) that captures the full request body, response body, and response headers for subsequent `/v1/chat/completions` calls. Recorded rows expose a red record-dot icon in the Messages log; clicking it opens a formatted modal showing Parameters, Conversation turns (role-coded), Tool calls, Reply, Usage pills, and Headers — JSON is a fallback only.

## Why

Users occasionally need to debug what their agent actually sent and received — prompts, tool calls, headers, streamed deltas. Before this change, only metadata (tokens, cost, model) was persisted; bodies were piped straight to the client and discarded.

## How

- **Schema**: `agents.record_messages boolean` (default false) + new `message_recordings` table (1:1 with `agent_messages`, `ON DELETE CASCADE`, all payload columns nullable JSONB). Only a cheap `recorded` boolean is added to `agent_messages` itself, with a partial index `WHERE recorded = true` for the "Recorded only" filter. Large payloads stay off the hot messages-list query path.
- **Capture**: a `CaptureSink` in `recording-capture.ts` with a 2 MB hard cap. `pipeStream` got a `writeOut()` wrapper so stream recording fires **post-transform** — Google/Anthropic routes record the OpenAI-compatible SSE the client actually saw, not upstream provider-native chunks.
- **Toggle**: extends `PATCH /api/v1/agents/:name` with `record_messages`. A new `AgentRecordingCacheService` (60s TTL) reads the flag per proxy call without a DB hit; the PATCH handler invalidates the cache so toggling off takes effect immediately.
- **Refactor**: extracted a generic `TtlFifoCache<K, V>` in `common/utils/` and rewrote both `AgentRecordingCacheService` and `TenantCacheService` on top of it.
- **API**: extended `GET /api/v1/messages/:id/details` with an optional `recording` field; new `DELETE /api/v1/messages/:id/recording` endpoint for per-row deletion (the message row stays, the payload is removed).
- **Frontend**: per-agent **Recording** card in Settings (with a pulsing "Recording" status pill when on), row indicator in `MessageTable`, "Recorded only" filter chip in `MessageLog`, new `RecordedMessageModal` with structured sections via `CodeBlock` / `HeadersTable` / `UsagePills`.
- **Security / privacy**: response auth headers (`set-cookie`, `authorization`, `proxy-authorization`) stripped at capture time. Recording content is explicitly excluded from the anonymous telemetry payload per the CLAUDE.md exclusion list.

## Test plan

- [ ] Backend unit: `npm test --workspace=packages/backend` — 4126 tests pass locally; new files at 100% line coverage (`recording-capture.ts`, `agent-recording-cache.service.ts`, `message-recording.service.ts`, `ttl-fifo-cache.ts`, `message-recording.entity.ts`).
- [ ] Backend e2e: `npm run test:e2e --workspace=packages/backend` — 125 tests pass.
- [ ] Frontend: `npx vitest run --workspace=packages/frontend` — 2337 tests pass.
- [ ] Shared: `npx jest --workspace=packages/shared` — 83 tests pass.
- [ ] `npx tsc --noEmit` clean in both `packages/backend` and `packages/frontend`.
- [ ] Manual: toggle on in Settings → send a chat completion → verify the row shows the red dot and the modal renders a formatted view (system/user/assistant turns, usage pills, headers table).
- [ ] Manual: toggle off → next call has no dot; existing recordings remain.
- [ ] Manual: delete a recording from the modal → dot disappears; message metadata row unchanged.
- [ ] DB: `SELECT id, recorded FROM agent_messages` and `SELECT message_id, size_bytes FROM message_recordings` reflect the flow.

---
## Summary by cubic
Adds an opt-in per-agent “Record messages” toggle (default off) that saves request body, response body, and response headers for chat completions. Recorded rows show a red dot in Message Log; clicking opens a structured modal for easy debugging.

- **New Features**
  - Per‑agent recording toggle in Settings; pulsing “Recording” status when on.
  - Captures request, post-transform SSE or JSON response, and response headers; 2 MB cap; strips `set-cookie`, `authorization`, and `proxy-authorization`.
  - Schema: `agents.record_messages`, `agent_messages.recorded` (+ partial index), and a new `message_recordings` table (1:1 with messages) to keep large payloads off the hot list path.
  - API: PATCH `/api/v1/agents/:name` accepts `record_messages`; GET `/api/v1/messages/:id/details` includes `recording`; DELETE `/api/v1/messages/:id/recording` removes payloads.
  - `AgentRecordingCacheService` (60s TTL) avoids per-call DB lookups; Message Log gets a “Recorded only” filter and a red-dot row button; new RecordedMessageModal shows Parameters, Conversation, Tool calls, Reply, Usage, and Headers.

- **Migration**
  - Run DB migrations.
  - No behavior change until recording is enabled per agent.
  - Enable via Settings → Recording or PATCH `/agents/:name` with `record_messages: true`.

<sup>Written for commit a63e67ec26cc087214e671a933f8bb0c6a69d84e. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/record-messages.md
+++ b/.changeset/record-messages.md
@@ -1,0 +1,5 @@
+---
+'manifest': minor
+---
+
+Add opt-in per-agent message recording. Toggle in Settings → Recording captures the full request body, response body, and response headers for subsequent proxy calls. Recorded rows show a record-dot icon in the Messages log; clicking it opens a formatted modal with Parameters, Conversation turns, Tool calls, Reply, Usage pills, and Headers. Payloads capped at 2 MB and kept out of the hot message-list query via a separate `message_recordings` table. Defaults off; never included in anonymous telemetry.
--- a/packages/backend/src/analytics/analytics.module.ts
+++ b/packages/backend/src/analytics/analytics.module.ts
@@ -6,6 +6,7 @@ import { Tenant } from '../entities/tenant.entity';
 import { LlmCall } from '../entities/llm-call.entity';
 import { ToolExecution } from '../entities/tool-execution.entity';
 import { AgentLog } from '../entities/agent-log.entity';
+import { MessageRecording } from '../entities/message-recording.entity';
 import { OtlpModule } from '../otlp/otlp.module';
 import { RoutingCoreModule } from '../routing/routing-core/routing-core.module';
 import { AggregationService } from './services/aggregation.service';
@@ -14,6 +15,7 @@ import { TimeseriesQueriesService } from './services/timeseries-queries.service'
 import { MessagesQueryService } from './services/messages-query.service';
 import { MessageDetailsService } from './services/message-details.service';
 import { MessageFeedbackService } from './services/message-feedback.service';
+import { MessageRecordingService } from './services/message-recording.service';
 import { SpecificityFeedbackService } from './services/specificity-feedback.service';
 import { AgentAnalyticsService } from './services/agent-analytics.service';
 import { OverviewController } from './controllers/overview.controller';
@@ -25,7 +27,15 @@ import { AgentAnalyticsController } from './controllers/agent-analytics.controll
 
 @Module({
   imports: [
-    TypeOrmModule.forFeature([AgentMessage, Agent, Tenant, LlmCall, ToolExecution, AgentLog]),
+    TypeOrmModule.forFeature([
+      AgentMessage,
+      Agent,
+      Tenant,
+      LlmCall,
+      ToolExecution,
+      AgentLog,
+      MessageRecording,
+    ]),
     OtlpModule,
     RoutingCoreModule,
   ],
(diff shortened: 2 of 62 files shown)


jagmarques:
Pre-generate `agent_messages.id` at request entry in the proxy controller and thread it into `RoutingMeta` so `buildMetaHeaders` can emit `X-Manifest-Message-Id` next to `X-Manifest-Tier`/`Model`/`Provider`. Today the id is allocated inside `buildMessageRow` (`uuid()` at `proxy-message-recorder.ts:67`) post-stream, so clients have no stable handle until they poll. About four lines: gen uuid early, add `messageId` to `RoutingMeta`, set the header in `buildMetaHeaders`, pass the same id into `buildMessageRow`. Unlocks correlation for external archivers and audit tools without polling; adjacent context in #1793.

PR #1698, opened 2026-04-23
feat: add Groq as first-party provider

guillaumegay13:
## Summary

Adds Groq as a first-party LLM provider across the full stack:

- **Shared registry**: New `groq` entry in `SHARED_PROVIDERS` with `gsk_` key prefix, OpenRouter prefix mapping, and `#f55036` brand color
- **Model discovery**: FetcherConfig hitting `api.groq.com/openai/v1/models` with non-chat filter for prompt-guard, safeguard, and orpheus models
- **Proxy routing**: ProviderEndpoint at `api.groq.com/openai` with OpenAI-compatible format, Bearer auth, and `stream_options.include_usage` support for token tracking
- **Frontend**: UI overlay tile with `Gq` initial, "Llama, Gemma, Mixtral — fast inference" subtitle, and API key URL pointing to `console.groq.com/keys`

## Pre-Landing Review

3 issues found, all auto-fixed:
- [CRITICAL] `provider-client.ts:41` — Missing `'groq'` in `SUPPORTS_USAGE_STREAM_OPTIONS` → added
- [INFO] `providers.spec.ts:12,95` — Stale "16" in test descriptions → updated to "17"
- [INFO] `package-lock.json` — Unrelated version bump → reverted

## Test Coverage

All new code paths covered:
- Groq model fetcher: parse response + non-chat filter + bearer auth header
- Groq proxy endpoint: base URL, format, path builder, headers
- Groq stream_options.include_usage injection for streaming requests
- Provider count assertions updated (16 → 17)
- Custom provider test example renamed from "Groq" to "Cerebras" to avoid collision

## Test plan
- [x] Backend tests pass (4100/4100)
- [x] Frontend tests pass (2299/2299)
- [x] Shared tests pass (83/83)

---
## Summary by cubic
Adds Groq as a first‑party provider with native model discovery, pricing sync, and OpenAI‑compatible routing. Users can connect `gsk_` keys, pick Groq models, and see correct Groq attribution with streaming usage tracked.

- **New Features**
  - Shared: added `groq` to `SHARED_PROVIDERS` with color `#F43E01`, `gsk_` key support (min length 50), no `openRouterPrefixes`; added Groq icon and API key URL.
  - Model discovery: queries `https://api.groq.com/openai/v1/models` with Bearer auth; filters non‑chat entries (all `compound*` variants, `prompt-guard`, `orpheus`); keeps `openai/gpt-oss-safeguard-20b`.
  - Pricing: added Groq to models.dev sync; falls back to known prices for Groq catalog (e.g., `llama-3.3-70b-versatile`, `qwen/qwen3-32b`).
  - Proxy + UI: routes via `https://api.groq.com/openai` (OpenAI format) and injects `stream_options.include_usage` for streaming; adds Groq tile (“Gq”, “Llama, Gemma, Mixtral — fast inference”); maps Free Models “Groq” to the first‑party tile.

- **Bug Fixes**
  - Proxy: preserve slash‑separated Groq model IDs when forwarding (e.g., `meta-llama/llama-guard-4-12b`).
  - Routing/UI: prefer the DB provider over prefix inference so Groq‑served `openai/*` models stay selectable and show the Groq provider.

<sup>Written for commit c745f8e67cd828ae2c22e6102d32005766ddb8d4. Summary will update on new commits.</sup>


SebConejo (changes requested):
@guillaumegay13  When adding groq as a provider, by default, a Compound model is displayed with with the logo of openrouter provider. See image. 

Image

This monel is not listed in the groq model list. But it is selected by default. So it's strange. 

What I undestood investing into it is that we receive 8 models from groq. But only 2 can receive price. So we display those 2 models. The price is not found from groq api, neither form models.dev but from openrouter.

The other 6 (compound-mini, llama-4-scout, qwen3-32b, etc.) are hidden because they have no pricing data, and the model picker filters out models with null prices (ModelPickerModal.tsx:39-40).

The automatic routing should chose between those 2 models. And we should make sure than those models are mentioned from Groq provider instead of OR

Additonnal note: 
Groq publishes all their prices at https://groq.com/pricing/. But not on the API. Any thoughts on how you'd want to handle this? An agent that periodically scrapes/checks provider pricing pages could keep things in sync automatically and allow us to show 6 other models from this provider.

Image

guillaumegay13:
@SebConejo Thanks for the careful review — pushed a fix commit (`9bf58338`) that addresses both findings. Branch is rebased on current `main` (the previous tip was 83 commits behind).

**Compound + OpenRouter logo bug**

Root cause was that the shared `groq` entry had `openRouterPrefixes: ['groq']`. That mapping made the OpenRouter pricing cache emit `groq/*` rows even when they weren't in the user's actual Groq `/models` response, which is how `compound` got selected by default *and* how it inherited the OR logo (the model entry literally came from the OR cache).

Fix: cleared `openRouterPrefixes` for Groq, and added a `^compound` filter to `PROVIDER_NON_CHAT.groq` so even if it shows up in `/models` later it doesn't get auto-assigned. Compound is a server-side router product, not a model the user picks directly.

Side fix: the original regex blocked `safeguard` (intended to catch `llama-prompt-guard`), which would have hidden `openai/gpt-oss-safeguard-20b` once it was on the Groq tile — that's a real chat model, so I narrowed the regex.

**6 hidden models due to missing pricing**

Manifest already has a three-tier pricing chain: **models.dev sync → OpenRouter cache → `known-model-prices.ts`** (last-resort hardcoded). Other providers without OR coverage (moonshot-v1, gemma-3-1b-it, gemini-pro-latest) already use this exact mechanism.

Fix: added Groq's documented catalog to `known-model-prices.ts`, verified against https://groq.com/pricing/ on 2026-05-01:

| Model | Input | Output |
|---|---|---|
| `openai/gpt-oss-20b` | $0.075 | $0.30 |
| `openai/gpt-oss-safeguard-20b` | $0.075 | $0.30 |
| `openai/gpt-oss-120b` | $0.15 | $0.60 |
| `meta-llama/llama-4-scout-…` | $0.11 | $0.34 |
| `meta-llama/llama-guard-4-12b` | $0.20 | $0.20 |
| `qwen/qwen3-32b` | $0.29 | $0.59 |
| `moonshotai/kimi-k2-instruct…` | $1.00 | $3.00 |
| `llama-3.3-70b-versatile` | $0.59 | $0.79 |
| `llama-3.1-8b-instant` | $0.05 | $0.08 |

(All per million tokens.) Last two also exist on OpenRouter so they're covered there too — listed here so a fresh install with no OR sync still gets accurate Groq pricing.

**On a periodic scraper for provider pricing pages**

Agree it'd be nice to have, but I'd keep it out of this PR — it's a separate moving piece (HTML diffing, ToS sensitivity, schedule). Happy to do it as a follow-up if you want; the existing `known-model-prices.ts` mechanism is what we already use for the same situation with Moonshot, Gemma, and Gemini Pro Latest, so this PR isn't introducing a new pattern.

All 4338 backend tests + 2610 frontend tests + 130 shared tests are green locally.

guillaumegay13:
Heads-up: while testing locally, `qwen/qwen3-32b` discovered from Groq still renders with the Qwen logo and Qwen-on-OR pricing ($0.08/$0.24 instead of Groq's $0.29/$0.59). That's a more general attribution bug that pre-dates this PR — `providerIdForModel` (`RoutingTierCard.tsx`) prefers prefix inference over the stored `m.provider`, and `ModelDiscoveryService.enrichModel` consults the OR cache before `known-model-prices.ts`. It affects every redistributor (Groq, Together, Cerebras-as-custom-provider, etc.), Groq just exposes it on every model.

PR #1772's `route.provider` is the right substrate to fix it. Doing that as a separate, focused follow-up PR — keeps this one scoped to "add Groq" and lets the attribution change get its own review and changeset (it silently changes pricing for moonshot-v1, gemma-3-1b-it, and gemini-pro-latest too). I'll link the follow-up here when it's up.

guillaumegay13:
Follow-up opened: #1788 (3-line frontend change + a small pricing-chain reorder). Standalone, doesn't block this PR.

SebConejo (changes requested):
Hey @guillaumegay13, the compound/OR root cause works. But I stil found two small things:

1. FreeModels.tsx:74: Groq maps to 'custom' but should be 'groq' now that it's first-party.

2. models-dev-sync.service.ts: Groq is missing from PROVIDER_ID_MAP, so models.dev pricing is skipped entirely. Adding groq: 'groq' fixes the missing prices and gives proper display names. By the way, if you add it, the ^compound regex in PROVIDER_NON_CHAT also needs updating to `(?:(?:^|\/|-)compound|prompt-guard|orpheus)` because models.dev returns groq/compound (with prefix).

So I'm seeing only 3 models, and 1 of them is without price.

Image


Could you also share screenshots of the model picker and routing page showing the expected result? It helps me to validate.

I tried the fix I'm talking about, And I got this result below:

Image

Image

guillaumegay13:
@SebConejo ready to test

SebConejo (changes requested):
Screenshot 2026-05-11 at 10 35 03
Screenshot 2026-05-11 at 10 35 08
Screenshot 2026-05-11 at 10 34 58

@guillaumegay13 I'm still seegin the same bugs. 

1. The logo of OR is displayed instead of grok
2. I can select only 2 models. 
3. I see a selected model that is not on this list of selectable models (GPT-OSS)

**How to reproduce**

From commit 8d71565:
1. I create my new agent
2. I connect only 1 provider: Groq with my gsk- key 
3. I activate the complexity default routing. 
4. I see those 3 errors mentioned above (Or logo instead of groq, only 2 models I can select, and the GPT-oss model pre-selected by manifest whereas I can decide to select it.

SebConejo:
### Root Cause Analysis: Three bugs, one root cause

All three issues stem from the same problem: **Groq's API returns model IDs with an `openai/` prefix** (e.g. `openai/gpt-oss-120b`), and the frontend's provider inference misidentifies them as OpenRouter models.

#### How it happens

The catch-all regex in `packages/shared/src/provider-inference.ts:20`:

```typescript
[/^[a-z][\w-]*\//, 'openrouter']
```

matches any `vendor/model` pattern and returns `"openrouter"`. So `openai/gpt-oss-120b` → `"openrouter"`.

This causes all three bugs:

**1. Only 2 models in the picker** — In `ModelPickerModal.tsx:121-134`, each model goes through `inferProviderFromModel(model_name)`. For `openai/gpt-oss-120b`, it returns `"openrouter"`. Then line 134 filters models by connected provider — the user connected **Groq**, not OpenRouter → the model is **filtered out** of the picker.

**2. OpenRouter logo instead of Groq** — `providerIdForModel()` in `RoutingTierCard.tsx:30` calls `inferProviderFromModel("openai/gpt-oss-120b")` → returns `"openrouter"` → renders the OpenRouter icon.

**3. GPT OSS 120B auto-assigned to tiers but not selectable** — The backend's tier auto-assign uses `getModelsForAgent()` which reads `cached_models` directly without any provider inference. It correctly sees GPT OSS 120B as a Groq model, scores it high on quality, and assigns it to Complex/Reasoning tiers. But the frontend picker filters it out as explained in bug 1.

#### Summary

The backend correctly stores `provider: "groq"` for these models. But the frontend overrides this with name-based inference, and that inference is wrong for Groq's `openai/`-prefixed model IDs — they get caught by the OpenRouter catch-all regex instead of being recognized as Groq models.

guillaumegay13:
@SebConejo can you please retest with the new local env skill and latest version ?

SebConejo (approved):

PR #1726, opened 2026-04-26
feat(analytics): add saved cost metric to Overview dashboard

SebConejo:
## Summary

- Add a "Saved cost" metric to the Overview dashboard ChartCard header, showing how much money routing saves compared to using a single model for all requests
- Baseline model auto-selected as the cheapest reasoning-capable model from current + historical providers, user-overridable via dropdown
- Full-page explainer modal detailing the calculation methodology (formula, auth type handling, limitations)

## What changed

**Backend:**
- `SavingsQueryService` with per-request savings calculation (single SQL aggregation query)
- `SavingsController` with 3 endpoints: GET /savings, PUT /savings/baseline, GET /savings/baseline-candidates
- `selectBaselineModel()` pure function for baseline auto-selection
- Migration adding `savings_baseline_model` nullable column to agents table
- Historical model support: models from past messages remain as baseline candidates even after disconnecting the provider

**Frontend:**
- "Saved cost" stat integrated into ChartCard header (4th column alongside Messages, Cost, Token usage)
- Baseline selector with dropdown (shows model name when selected, full pricing in dropdown list)
- Full-page SavingsExplainer modal (replaces Overview content, back button to return)
- `Select` component: added optional `displayValue` prop for custom trigger text

## Test plan

- [ ] Backend: 234 test suites pass (4286 tests), including new savings service, controller, DTO, and migration specs
- [ ] Frontend: 129 test suites pass, including SavingsCard component tests and updated Overview tests
- [ ] Manual: verify savings card renders on Overview with connected providers
- [ ] Manual: verify baseline dropdown changes the savings calculation
- [ ] Manual: verify explainer modal opens/closes correctly

---
## Summary by cubic
Adds a “Saved cost” metric and a savings chart to the Overview, showing how much routing saves vs the most expensive routed‑model baseline. The baseline is computed per request at ingest, and savings only sum positive deltas.

- **New Features**
  - Savings API: GET `/savings` (totals, auth-type breakdown) and GET `/savings/timeseries` (hourly/daily bars).
  - Auto baseline: pick the most expensive model selected in routing (tiers, specificity, header tiers, fallbacks) at API‑key rates; enrich subscription/local models with real API pricing and capabilities.
  - Per-request baseline at ingest: store `baseline_model_id` and `baseline_cost_usd`; totals use stored values; per-request savings are clamped to $0 and only positives sum.
  - Overview UI: “Savings” stat is clickable like cost/tokens/messages; `SavingsChart` shows single savings‑per‑interval bars; info tooltip with a “More details” link; explainer simplified and clarifies logic per access type (API key, subscription, local). Manual baseline dropdown removed (always auto mode).

- **Bug Fixes**
  - Stabilized tests with `vi.waitFor`, isolated savings API mocks, and caught fetch errors.

<sup>Written for commit 2f652230efc135dfa410e666a6eb76e51ec6bc5d. Summary will update on new commits. <a href="https://cubic.dev/pr/mnfst/manifest/pull/1726?utm_source=github">Review in cubic</a></sup>


brunobuddy (dismissed):
Choosing a middle-market model as a fixed baseline for everyone does not make sense. 

Saved costs for each agent should be based on their consumption without Manifest, not a fixed model for everyone.

SebConejo:
Screenshot 2026-04-27 at 15 12 07
Screenshot 2026-04-27 at 15 12 06
screencapture-localhost-16860-agents-mowgly-2026-04-27-15_12_13

brunobuddy (dismissed):
The UX and messages are still hard to understand:
- The "baseline" term is ambiguous and not necessary. You could replace by "the most expensive model on your routing setup" as you already do in the page
- "Saved cost" is unnatural => Replace by "savings" or "Cost saved"
-  The modal text is unnatural: "Savings compare what you paid against what the most expensive model in your routing setup (at API key rates) would have cost for the same request." => Example of natural way of saying that "We calculate savings by comparing your actual cost to what the request would have cost on the most expensive model in your routing setup"
- The "vs" dropdown is amiguous and an unnecessary complication. I was expecting the chart to refresh. IMO You can just drop it. 
- Saving page is too long and can be shrinked: the baseline section can be removed entierly. The baseline principle is already explained on the next paragraph. the "subscription fees" bullet point at the bottom is already explained above. You also can remove the token line from the tables and many other things to simplify.

In few words, this could look way more simpler that how it is right now

SebConejo:
> The UX and messages are still hard to understand:
> 
> * The "baseline" term is ambiguous and not necessary. You could replace by "the most expensive model on your routing setup" as you already do in the page
- I find this proposal worst. It's too long. "Baseline" names the concept (a defined comparison point) rather than the calculation. We already explain tis definition. Its name should not be how it is defined. I think its is very fine to start with and and move forward. 

> ✅ * "Saved cost" is unnatural => Replace by "savings" or "Cost saved"
> ✅ * The modal text is unnatural: "Savings compare what you paid against what the most expensive model in your routing setup (at API key rates) would have cost for the same request." => Example of natural way of saying that "We calculate savings by comparing your actual cost to what the request would have cost on the most expensive model in your routing setup"
> * The "vs" dropdown is amiguous and an unnecessary complication. I was expecting the chart to refresh. IMO You can just drop it.
- I removed it but and add a saving bar chart. The chart will be weird the first month because the daily baseline cost will exist only after the deployment.

> ❌ * Saving page is too long and can be shrinked: the baseline section can be removed entierly. The baseline principle is already explained on the next paragraph. the "subscription fees" bullet point at the bottom is already explained above. You also can remove the token line from the tables and many other things to simplify.
- I don't agree. I really think this is important to provide this text as it is. I'm not interesting to show graphs and numbers without those explanations. I also think that if people don't care about this information, they just will never go until this modal. So the UX is adapted for them. 

Removing context and explanation isn't simplification. it makes things less clear. We can revisit the placement in the future when we provide an space with documentation. But for now I think we should move forward instead of polishing something that's already working well.
 
> In few words, this could look way more simpler that how it is right now

PR #1775, opened 2026-04-30
fix(agents): clean up data on delete so recreated agents start fresh

### LINKED ISSUE #1765: Recreating an agent with a deleted agent's name brings back old data
## What happened

When you delete an agent and create a new one with the same name, the new agent inherits the deleted agent's data. Messages, routing config, and history from the deleted agent all reappear. The new agent is not a clean slate.

## Expected behavior

Creating an agent with a previously used name should produce a fresh agent with no data. The user should not see messages, routing config, or any other state from the deleted agent.

## Steps to reproduce

1. Create an agent named "test-agent"
2. Send a few messages through it
3. Delete the agent
4. Create a new agent named "test-agent"
5. The old messages and routing config are back

## Environment

| | |
|---|---|
| Mode | Cloud (not tested in local yet) |
| Platform | macOS (Apple Silicon) |

## Additional context

The underlying data doesn't need to be hard-deleted. The issue is purely UX: when a user creates an agent with a name they used before, they expect a blank agent. The simplest fix might be to generate a new internal ID on creation rather than reusing the old record.

## Screenshots

https://github.com/user-attachments/assets/1a7047ec-94c2-4601-95c8-2a9afdaae33d

brunobuddy:
### 💭 Why
Recreating an agent with a previously used slug pulled back the deleted agent's messages, costs, and routing config. The agents row was the only thing being deleted, so anything keyed by `tenant_id+agent_name` (or by the in-memory resolve cache) re-attached itself.

### ✨ What changed
- `deleteAgent` now hard-clears denormalised rows (`agent_messages`, `agent_logs`, `notification_rules`/`notification_logs`) and agent-id-scoped tables (tier/specificity/header/user-provider, `llm_calls`, `tool_executions`) in one transaction.
- Invalidates `ResolveAgentService` and `RoutingCacheService` after delete so a recreated slug gets fresh routing state.
- Cleanup migration purges existing orphan rows on upgrade.
- E2E spec reproduces the bug end-to-end and asserts the recreated agent is empty across `/messages`, `/costs`, and `/routing/:name/tiers`.

### 👤 For users
Deleting an agent and creating a new one with the same name now produces a clean slate. Fixes #1765.

### 🔧 For operators
The new migration `1782200000000-CleanupOrphanedAgentRows` runs on boot and removes orphan rows left behind by previous deletes. Snapshot tables with no TypeORM entity are guarded with `to_regclass`, so the migration is safe across schema variants.

### DIFF
--- a/.changeset/agent-recreate-clean-slate.md
+++ b/.changeset/agent-recreate-clean-slate.md
@@ -1,0 +1,5 @@
+---
+"manifest": patch
+---
+
+Recreating an agent with a deleted agent's name no longer surfaces the deleted agent's messages, costs, or routing config. Deletion now hard-clears denormalised rows and invalidates the resolve/routing caches.
--- a/packages/backend/src/analytics/services/agent-lifecycle.service.spec.ts
+++ b/packages/backend/src/analytics/services/agent-lifecycle.service.spec.ts
@@ -4,13 +4,16 @@ import { ConflictException, NotFoundException } from '@nestjs/common';
 import { DataSource } from 'typeorm';
 import { AgentLifecycleService } from './agent-lifecycle.service';
 import { Agent } from '../../entities/agent.entity';
+import { ResolveAgentService } from '../../routing/routing-core/resolve-agent.service';
+import { RoutingCacheService } from '../../routing/routing-core/routing-cache.service';
 
 describe('AgentLifecycleService', () => {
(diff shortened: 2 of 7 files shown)


brunobuddy:
Closing in favor of a soft-delete implementation that preserves analytics data. New PR will follow on fix/agent-recreate-soft-delete.

PR #1788, opened 2026-05-01
fix(routing): attribute models by connection's provider, not model-id prefix

### LINKED ISSUE #1789: Routing attribution: redistributed models render with the wrong provider logo and price
## Summary

The routing UI derives a model's logo (and the discovery pipeline derives its price) from the **prefix of the model id**, not from the **provider that actually serves it**. That breaks for any provider whose catalog includes other vendors' model ids — most visibly Groq, but also Together, Cerebras-as-custom-provider, etc.

## Reproduce

1. Connect a Groq provider with a real \`gsk_\` key.
2. Open Routing → Default tab.
3. Auto-assignment picks a Groq-served model whose id has a vendor prefix (e.g. \`qwen/qwen3-32b\`).

**Observed:**
- Logo: orange Qwen swirl.
- Price: ~\$0.080 in / \$0.240 out per 1M (Qwen-on-OpenRouter pricing).

**Expected:**
- Logo: Groq's lightning bolt.
- Price: \$0.290 in / \$0.590 out per 1M (Groq's published price for that model — https://groq.com/pricing).

## Root cause

Two prefix-based heuristics fire even when better information is available:

1. **`packages/frontend/src/pages/RoutingTierCard.tsx` → `providerIdForModel`**
   When a model row has a stored \`provider\` (\"groq\"), the code intentionally **prefers** the prefix-inferred provider over the stored one. The comment says it's there to avoid an Ollama colon-suffix heuristic mis-routing to local Ollama, but the precedence is now too aggressive — every redistributor loses.

2. **`packages/backend/src/model-discovery/model-discovery.service.ts` → `enrichModel`**
   The pricing chain is currently \`models.dev → OpenRouter cache → known-model-prices.ts\` (last-resort). When OR has the same model id under a different (cheaper) inference provider, OR wins, and a connection's reported pricing reflects that provider's catalog instead of the connection's own.

PR #1772 already introduced the right substrate (`route.provider` carrying the connection's actual provider per row) — both call sites just need to use it.

## Proposed fix

- **Frontend:** in `providerIdForModel`, when the stored \`m.provider\` resolves to a registered first-party provider, return it. OpenRouter remains the documented exception (its rows really do represent vendor-prefixed models served on behalf of those vendors). Ollama still resolves via its dbId (was a special case before, becomes a general rule).
- **Backend:** consult `known-model-prices.ts` **before** models.dev / OR. Curated entries become authoritative — a behaviour change for moonshot-v1, gemma-3-1b-it, gemini-pro-latest, but it matches their original intent.

## Tracked by

PR #1788 (proposed fix, end-to-end verified locally with a Groq connection serving \`qwen/qwen3-32b\`).

Surfaced while landing #1698 (add Groq as a first-party provider).

guillaumegay13:
Closes #1789. Builds on #1772.

## Summary

The routing UI derived a model's logo from the prefix of its model id, which broke for any provider that redistributes other vendors' models. Most visibly, a Groq connection serving `qwen/qwen3-32b` rendered with the Qwen logo and Qwen-on-OpenRouter pricing (~$0.08/$0.24 per 1M) instead of Groq's pricing ($0.29/$0.59).

See #1789 for the full bug report.

## Changes

**Frontend — `RoutingTierCard.providerIdForModel`**

When a model row has a stored `provider` that resolves to a registered first-party provider, that wins over prefix inference. OpenRouter remains the documented exception because its rows really do represent vendor-prefixed models served on behalf of those vendors. Ollama still resolves via `dbId` (was a special case before, is now the general rule).

**Backend — `ModelDiscoveryService.enrichModel`**

`known-model-prices.ts` is now consulted **before** models.dev and the OpenRouter cache. Curated per-provider prices win over upstream catalogs that may attribute the same model id to a different (cheaper) inference provider.

This is a **behaviour change** for the existing curated entries (moonshot-v1-*, gemma-3-1b-it, gemini-pro-latest): they become authoritative instead of last-resort. In practice all three already worked because the upstream sources lacked the data — this just locks in the intent. The `known-model-prices.ts` docstring is updated accordingly.

When known-prices wins on pricing we still consult models.dev for capability flags (reasoning, tool-call), mirroring the existing price-already-set branch — those drive tier auto-assignment scoring and shouldn't be lost.

## Why this is a separate PR from #1698

#1698 is scoped to "add a provider." This bug affects every redistributor (Groq, Together, Cerebras-as-custom-provider…) and changes pricing precedence globally — different concern, separate review and changeset.

## Test plan
- [x] Backend tests pass (4326/4326) — 5 new in `model-discovery.service.spec.ts` covering: known-prices wins over models.dev / OR; capability flags still merged when known-prices wins; non-curated models still fall through to upstream sources.
- [x] Frontend tests pass (2614/2614) — 4 new in `RoutingTierCard.test.tsx` covering: stored provider wins over prefix for first-party; OpenRouter prefix-wins exception preserved; unknown prefix still resolves via dbId; Ollama-style behaviour preserved.
- [x] Shared tests pass (129/129).
- [x] Verified locally: a Groq connection serving `qwen/qwen3-32b` now shows the Groq logo and Groq's pricing.
- [x] End-to-end verified: OpenAI Python SDK → `model="auto"` → routed to `qwen/qwen3-32b` via Groq → response back, usage telemetry recorded.

### DIFF
--- a/.changeset/route-provider-attribution.md
+++ b/.changeset/route-provider-attribution.md
@@ -1,0 +1,27 @@
+---
+"manifest": patch
+---
+
+fix(routing): attribute models by their connection's provider, not by model-id prefix
+
+The routing UI used to derive a model's logo from the prefix of its model id,
+which broke for any provider that redistributes other vendors' models. Most
+visibly, a Groq connection serving `qwen/qwen3-32b` rendered with the Qwen
+logo and Qwen-on-OpenRouter pricing (≈$0.08/$0.24 per 1M) instead of Groq's
+own pricing ($0.29/$0.59).
+
+Two changes:
+
+- **Frontend** (`RoutingTierCard.providerIdForModel`): when a model row has a
+  stored `provider` that resolves to a registered first-party provider, that
+  wins over prefix inference. OpenRouter remains the documented exception
+  because its rows really do represent vendor-prefixed models served on behalf
+  of those vendors.
+- **Backend** (`ModelDiscoveryService.enrichModel`): `known-model-prices.ts`
+  is now consulted *before* models.dev and the OpenRouter cache, so curated
+  per-provider prices win over upstream catalogs that may attribute the same
+  model id to a different (cheaper) inference provider. Behaviour change for
+  the existing entries (moonshot-v1, gemma-3-1b-it, gemini-pro-latest): they
+  become authoritative instead of last-resort, which matches their intent.
+
+Builds on #1772, which introduced `route.provider` as the routing identity.
--- a/packages/backend/src/model-discovery/known-model-prices.ts
+++ b/packages/backend/src/model-discovery/known-model-prices.ts
@@ -1,8 +1,20 @@
 /**
- * Last-resort hardcoded pricing for models that no external pricing source covers.
+ * Hand-curated, per-provider authoritative pricing.
  *
- * This is used ONLY when both models.dev and OpenRouter have no data for a model.
- * Keep this list minimal — prefer upstream sources. Prices are per-token (not per-million).
+ * **Priority:** entries here win over models.dev and the OpenRouter cache
+ * during model discovery. The reason: the same model id can exist on
+ * multiple inference providers at different prices (e.g. `qwen/qwen3-32b`
+ * appears in Qwen's OR listing at one price and is also served by Groq at
+ * a different price), and a connection's reported pricing must reflect
+ * *that connection's provider*, not whatever is cheapest in upstream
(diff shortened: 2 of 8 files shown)


guillaumegay13:
Fixed in 4d5734e — wrapped the known-prices result in `applyCapabilities()` so models.dev reasoning/tool-call flags still apply when curated pricing wins, mirroring the existing price-already-set branch. Added a test locking in the capability-merge behaviour.

PR #1810, opened 2026-05-04
feat: add /v1/messages Anthropic-compatible endpoint

guillaumegay13:
### ✨ What changed
- New `POST /v1/messages` mirrors the existing `/v1/responses` pattern. A small adapter translates Anthropic Messages to chat-completions for routing, then back on the response side.
- DeepSeek `reasoning_content` round-trips through Anthropic `thinking` blocks so tool-use loops survive.
- Anthropic SDK and Claude Code added to the agent picker. Claude Code's setup card shows a paste-ready `~/.claude/settings.json` block, no shell tooling.

### 💭 Why
Anthropic-SDK clients (Claude Code, `@anthropic-ai/sdk`) couldn't point `ANTHROPIC_BASE_URL` at Manifest. Now they can, and they go through the same tier scoring, fallback chain, and recording as every other proxy path.

### 👤 For users
Drop the env block into `~/.claude/settings.json`, run `claude`, watch requests land tier-routed in the Manifest dashboard.

### 🔧 For operators
No migrations, no env vars, no flags. The endpoint is gated by the existing `AgentKeyAuthGuard`.

### 📝 Notes
- 100% line coverage on every new/modified file in `routing/proxy/`.
- 4352 backend unit + 138 e2e + 2617 frontend + 129 shared tests pass.
- End-to-end verified with the Anthropic Python SDK: tool-use round-trip, streaming text + tool, multi-turn, system prompts, stop_sequences, 10x concurrent.
- Streaming `reasoning_content` is bridged in the non-stream path only. SSE delta translation is a follow-up.

---
## Summary by cubic
Adds an Anthropic Messages-compatible endpoint at POST `/v1/messages` so Anthropic SDK clients (Claude Code, `@anthropic-ai/sdk`) can use Manifest via `ANTHROPIC_BASE_URL`. Requests translate to internal chat-completions for routing and convert back on the response side, including SSE, tools, images, and reasoning traces.

- **New Features**
  - New `POST /v1/messages` with an adapter that maps Anthropic Messages ⇄ chat-completions; streams are translated into Anthropic `message_*` SSE events.
  - Bridges `reasoning_content` via Anthropic `thinking` blocks so DeepSeek and similar tool-use loops round-trip.
  - Dashboard adds Anthropic SDK and Claude Code options with a paste-ready `~/.claude/settings.json`; Anthropic snippets strip a trailing `/v1` from the base URL.

- **Bug Fixes**
[...]

PR #1811, opened 2026-05-05
fix: preserve reasoning_content in multi-turn tool-use for DeepSeek, Kimi K2, and OpenCode Go

### LINKED ISSUE #1729: Support per-provider extra_body to disable DeepSeek thinking mode
## Problem

DeepSeek models (V4 Flash, V4 Pro) have **thinking mode enabled by default** via the DeepSeek API. There is currently no way to disable this in Manifest.

When Manifest routes requests to a DeepSeek model, the API returns `reasoning_content` that consumes the majority of allocated tokens. The actual `content` field is often **empty or severely truncated**  -  especially when `max_tokens` is below 4000.

### Real-world impact (self-hosted Manifest)

- DeepSeek V4 Flash routed as `standard` tier fallback → messages returned empty content
- DeepSeek V4 Pro routed as `complex` tier → reasoning_content ate all tokens, content was empty
- Had to **completely remove DeepSeek models from all 4 routing tiers** as a workaround

### Example of what happens

Simple `standard` tier request with DeepSeek V4 Flash:

- max_tokens: 4096
- reasoning_content: ~3800 tokens consumed
- content: empty or `< 100` tokens
- Result: user gets a blank response

## Root cause

The DeepSeek API accepts `"thinking": {"type": "disabled"}` in the request body to turn off thinking mode. But Manifest has **no mechanism to inject `extra_body` parameters per provider**.

The workaround would be:

```json
{
  "model": "deepseek-chat",
  "messages": [...],
  "thinking": {"type": "disabled"}
}
```

## Proposed solution

Add a per-provider `extra_body` field in the Manifest provider configuration that gets merged into the request body sent to the provider API.

This would allow:

- DeepSeek: `{"thinking": {"type": "disabled"}}`
- Other providers: any future provider-specific body parameters

### Possible UI/config approaches

1. **Provider-level setting**  -  in the provider connection form, a textarea for "Extra body (JSON)" merged into every request
2. **Per-model setting**  -  in the routing tier, an optional JSON field per model
3. **Both**  -  provider default + per-model override

## Affected versions

- Manifest: self-hosted `manifestdotbuild/manifest:latest`
- DeepSeek API (api.deepseek.com)
- All DeepSeek models (V4 Flash, V4 Pro, etc.)

## Workaround used

Completely removed DeepSeek models from all routing tiers and replaced with alternatives (Claude Haiku 4.5, Qwen 3.5 Plus, GPT-5.5, Ollama local models).

## Additional error: reasoning_content must be passed back

In multi-turn conversations, DeepSeek API throws:

```
{"error":{"message":"The reasoning_content in the thinking mode must be passed back to the API.","type":"invalid_request_error"}}
```

This means Manifest would also need to store reasoning_content from DeepSeek responses and pass it back in subsequent requests. This adds unwanted complexity. **Disabling thinking mode via extra_body would avoid both issues.**

flowing-abyss:
## Problem

When a reasoning model (DeepSeek, Kimi K2, or similar) responds with both `reasoning_content` and `tool_calls` in the same assistant turn, the next request in the conversation fails with:

> The reasoning_content in the thinking mode must be passed back to the API.

This happens because OpenAI-compatible clients and SDKs treat `reasoning_content` as an unknown field and strip it when they replay the conversation history for the next turn. These providers require the field to be present in every assistant message that had reasoning enabled — so the stripped message causes a hard rejection.

Fixes #1729.

## Solution

This PR applies the **cache-and-reinject pattern** that Manifest already uses for two analogous cases:

- `ThinkingBlockCache` — Anthropic extended thinking blocks
- `ThoughtSignatureCache` — Google thought signatures

A new `ReasoningContentCache` (30-minute TTL, lazy cleanup) stores the `reasoning_content` string from the provider response, keyed by `sessionKey + firstToolCallId`. On the next request, `sanitizeOpenAiMessages` looks up the cache and re-injects the field into any assistant message that has `tool_calls` but is missing `reasoning_content`.

The fix is gated behind `supportsReasoningContent(endpointKey, model)` — currently true for:
- `deepseek` (native endpoint)
- `moonshot` (native endpoint — Kimi K2)
- `opencode-go` — Kimi K2 (`opencode-go/kimi-*`) and DeepSeek (`opencode-go/deepseek-*`) models
- `openrouter` + any model whose name starts with `deepseek/` or `moonshotai/`

All other providers are completely unaffected.

## What changed

| File | Change |
|------|--------|
| `reasoning-content-cache.ts` | New TTL cache class (mirrors `ThinkingBlockCache`) |
| `provider-client-converters.ts` | `supportsReasoningContent()`, `createDeepSeekStreamTransformer()`, re-injection logic in `sanitizeOpenAiMessages` |
| `proxy-response-handler.ts` | Streaming branch + `cacheReasoningContent()` for non-streaming responses |
| `proxy-types.ts` | New `ReasoningContentLookup` type added to `ForwardOptions` |
| `provider-client.ts` | Passes `reasoningContentLookup` through `buildRequest`; exposes `createDeepSeekStreamTransformer` as a property |
| `proxy-fallback.service.ts` | Threads `reasoningContentLookup` through the fallback chain |
| `proxy.service.ts` | Creates the lookup closure from the cache; passes it to the forward and fallback paths |
| `proxy.module.ts` | Registers `ReasoningContentCache` as a provider |
| `proxy.controller.ts` | Injects `ReasoningContentCache`; passes it to stream and non-stream response handlers |
| `__tests__/reasoning-content-cache.spec.ts` | Full unit test suite for the new cache |
| `__tests__/provider-client-converters.spec.ts` | Tests for `supportsReasoningContent`, `createDeepSeekStreamTransformer`, and re-injection logic |
| `__tests__/proxy-response-handler.spec.ts` | Tests for the streaming branch and `cacheReasoningContent` |
| `__tests__/proxy.service.spec.ts` / `proxy.controller.spec.ts` | Updated constructors to include the new cache mock |

## Relationship to PR #1810

PR #1810 addresses the Anthropic endpoint path. This PR is independent and covers the OpenAI-compatible path — DeepSeek native, Kimi K2 native, OpenCode Go (Kimi K2 and DeepSeek models), and all of the above via OpenRouter.

### DIFF
--- a/packages/backend/src/routing/proxy/__tests__/provider-client-converters.spec.ts
+++ b/packages/backend/src/routing/proxy/__tests__/provider-client-converters.spec.ts
@@ -1,4 +1,8 @@
-import { sanitizeOpenAiBody } from '../provider-client-converters';
+import {
+  sanitizeOpenAiBody,
+  supportsReasoningContent,
+  createDeepSeekStreamTransformer,
+} from '../provider-client-converters';
 
 describe('provider-client-converters', () => {
   describe('sanitizeOpenAiBody', () => {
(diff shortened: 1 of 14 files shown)


guillaumegay13:
hey @flowing-abyss I made a fix to preserve reasoning_content especially for DeepSeek & OpenCode Go (not yet direct Kimi, I'll have a look) here https://github.com/mnfst/manifest/pull/1902 so I expect it to work now.

I did not implement your new caching strategy yet for reasoning_content, will check further.

flowing-abyss:
Thanks for #1902! That fix handles the case where a client correctly echoes `reasoning_content` back — Manifest no longer strips it.

The failure I'm solving is different: most OpenAI-compatible SDKs (Go, Python, TypeScript) silently **drop** `reasoning_content` from the assistant message when they replay conversation history. So even with #1902 merged, DeepSeek still rejects turn N+1 with "The reasoning_content in the thinking mode must be passed back to the API" — because the field is absent before the request even reaches Manifest.

This PR adds a cache-and-reinject layer to handle that gap, using the same pattern already in place for `ThinkingBlockCache` (Anthropic extended thinking) and `ThoughtSignatureCache` (Google):

1. After each assistant turn, a `ReasoningContentCache` stores the accumulated `reasoning_content`, keyed by the first `tool_call.id` (extracted from the SSE stream or the non-streaming JSON response).
2. On the next turn, `sanitizeOpenAiMessages` checks for missing `reasoning_content` on an assistant message that has `tool_calls`, and re-injects the cached value.

I've rebased onto current `main` (post-#1902), resolved all conflicts, and merged the `supportsReasoningContent` implementations — keeping your regex-based endpoint guard and adding `moonshot` native endpoint and `moonshotai/` OpenRouter prefix (Kimi K2). All tests pass.

guillaumegay13:
thanks @flowing-abyss, very clear! adding @brunobuddy as reviewer to approve the PR.

PR #1845, opened 2026-05-07
refactor(oauth): extract RedirectPkceOauthBaseService

guillaumegay13:
### ✨ What changed
- New \`RedirectPkceOauthBaseService\` owns the loopback callback server, PKCE state, token exchange/refresh, and revoke logic.
- \`OpenaiOauthService\` becomes a 16-line subclass that supplies the OpenAI config; public API and test contract preserved.
- Config supports optional \`client_secret\` and \`extraAuthorizeParams\` so providers like Google can plug in without further base changes.

### 💭 Why
A second redirect-PKCE provider (Gemini, gemini-cli flow) would otherwise duplicate ~290 lines of identical OAuth boilerplate.

### 📝 Notes
- All 55 OpenAI OAuth specs pass without changes; full OAuth + proxy suite (1388 tests) green.
- Internal-only refactor, no user-visible impact, empty changeset attached.

---
## Summary by cubic
Extracted a reusable redirect-PKCE OAuth base and fixed token unwrap when providers omit a refresh token. `OpenaiOauthService` now extends the base; no public API change.

- **Refactors**
  - Added `RedirectPkceOauthBaseService` to handle loopback callback server, PKCE state, token exchange/refresh, revoke, and unwrap.
  - Simplified `OpenaiOauthService` to a thin subclass that supplies provider config (id, URLs, scope, callback port); config supports optional `client_secret` and `extraAuthorizeParams`.

- **Bug Fixes**
  - Unwrap returns a valid access token when `refresh_token` is absent and unexpired; returns null if expired with no refresh. Added regression tests for both cases.

<sup>Written for commit c140de7954dce6dcca875939330b5ce2a31c5cd6. Summary will update on new commits. <a href="https://cubic.dev/pr/mnfst/manifest/pull/1845?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/.changeset/puny-stars-laugh.md
+++ b/.changeset/puny-stars-laugh.md
@@ -1,0 +1,7 @@
+---
+---
+
+Internal: extract a shared `RedirectPkceOauthBaseService` from the OpenAI
+OAuth service so future redirect-PKCE providers reuse the loopback
+callback server, PKCE state, token exchange/refresh, and revoke logic.
+No behavior change.
--- a/packages/backend/src/routing/oauth/openai-oauth.service.spec.ts
+++ b/packages/backend/src/routing/oauth/openai-oauth.service.spec.ts
@@ -230,6 +230,21 @@ describe('OpenaiOauthService', () => {
       fetchMock.mockRejectedValue(new Error('network'));
       expect(await svc.unwrapToken(JSON.stringify(blob), 'a', 'u')).toBeNull();
     });
+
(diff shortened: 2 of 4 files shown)


guillaumegay13:
@cubic-dev-ai fixed in 509e80541 — moved `!blob.r` check after the expiry short-circuit and added two regression tests covering valid-no-refresh and expired-no-refresh cases.

PR #1857, opened 2026-05-07
fix(routing): show params button when DeepSeek is a fallback

guillaumegay13:
### ✨ What changed
- \`RoutingTierCard\` visibility for the Model Parameters button now scans \`[primary, ...fallback_routes]\` for a params-compatible provider.
- Dialog's provider-default hint follows the first compatible route in that ordered list.
- Two new test cases: DeepSeek-as-fallback shows the icon; OpenAI + Anthropic mix hides it.

### 💭 Why
Today the icon hides if the primary route is OpenAI but DeepSeek is a fallback, even though the proxy already applies tier-level \`param_defaults\` per attempt against whichever provider an iteration actually targets (\`filterParamDefaultsForProvider\`). The button should be reachable whenever any route in the slot can consume a known param key.

### 📝 Notes
- No backend change.
- Header-tier card and unsupported-provider visibility are tracked as separate follow-ups.

---
## Summary by cubic
Show the Model Parameters button whenever any route in a tier—primary or fallback—supports params (e.g., DeepSeek). The button appears next to that route; the chip hides when only a fallback is compatible, and the dialog hint uses the first compatible provider.

- **Bug Fixes**
  - Fallback-row visibility now resolves the provider ID from `fallback_routes` or the model catalog, so it works without structured routes; tests cover both the fallback iteration and catalog-resolved paths.

<sup>Written for commit 80d38ad997479a9041c379356c40c317c1681120. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/params-fallback-visibility.md
+++ b/.changeset/params-fallback-visibility.md
@@ -1,0 +1,5 @@
+---
+'manifest': patch
+---
+
+Show the Model Parameters button on a routing tier when any route in the tier — primary or fallback — uses a params-compatible provider. Previously the button was gated to the primary route only, so a tier with DeepSeek configured as a fallback hid the toggle even though the proxy already applies tier-level `param_defaults` to whichever provider an attempt actually targets. The dialog's provider-default hint follows the first compatible route in the ordered (primary, …fallbacks) list.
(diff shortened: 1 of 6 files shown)


SebConejo (changes requested):
Unit tests pass, but the feature is not working in any real scenario. The button is invisible in both primary and fallback with DeepSeek.

SebConejo (approved):
We saw it together and validated the UI. We could not repdoduce the former bugs I mentioned

PR #1865, opened 2026-05-08
feat: add Nanobot to supported personal AI agents

guillaumegay13:
### ✨ What changed
- Adds `nanobot` to `AGENT_PLATFORMS` and `PLATFORMS_BY_CATEGORY.personal` in `manifest-shared`.
- New `NanobotSetup` component renders a paste-ready `~/.nanobot/config.json` block.
- New tab in `SetupStepAddProvider` between Hermes and Claude Code, plus filtered-mode match.
- Ships `/icons/nanobot.png` (cat head cropped from upstream `images/nanobot_logo.png`).
- Tests for the snippet helper, the new component, and the expanded tab strip.

### 💭 Why
Nanobot ([HKUDS/nanobot](https://github.com/HKUDS/nanobot)) is an OpenAI-compatible personal AI assistant. Surfacing it in the agent picker lets users pick it during onboarding and get the right setup snippet without leaving the wizard.

### 👤 For users
Nanobot now appears as a personal AI agent option. Selecting it shows a JSON block to drop into `~/.nanobot/config.json` that points the `manifest` provider at the proxy and selects model `auto`.

### 📝 Notes
Validated locally against `MANIFEST_MODE=local` build. Shared 184/184, frontend 2816/2816, backend telemetry+DTO 62/62, `tsc --noEmit` clean on both packages, lint 0 errors.


guillaumegay13:
Tested this locally with Nanobot and updated the setup snippet accordingly.

The key issue was that Nanobot does not support an arbitrary provider name like `manifest`. The working config uses Nanobot's built-in `custom` provider pointing at Manifest's OpenAI-compatible endpoint:

```json
{
  "providers": {
    "custom": {
      "apiKey": "mnfst_...",
      "apiBase": "http://127.0.0.1:38265/v1",
      "extraHeaders": null,
      "extraBody": null
    }
  },
  "agents": {
    "defaults": {
      "provider": "custom",
      "model": "auto"
    }
  }
}
```

I also updated the UI copy to make it clear these keys should be merged into an existing `~/.nanobot/config.json`, not treated as a standalone arbitrary provider block.

PR #1879, opened 2026-05-11
fix: preserve Anthropic cache tokens through /v1/messages

### LINKED ISSUE #1871: /v1/messages requests with cache_control markers don't trigger Anthropic prompt caching
Have been testing the recently merged https://github.com/mnfst/manifest/pull/1810 (thanks @guillaumegay13). So far so good - the only issue I've run into is around cache_control.

Specifically - Anthropic Messages requests sent to `POST /v1/messages` with `cache_control: { type: "ephemeral" }` result in **zero cache hits** when forwarded to Anthropic. Should round-trip through the proxy so Anthropic creates and reads the cache normally.

Manifest's `agent_messages` table records `cache_creation_input_tokens = 0` and `cache_read_input_tokens = 0` on every request through this path, even when the request body contains a properly-shaped cache_control marker on a >2048-token system block.

## Repro

`manifest@6.1.0` self-hosted (Docker, default config). Anthropic provider configured with an api-key auth.

```bash
KEY=mnfst_<your-agent-key>

# Build a >2048-token system prompt to exceed Haiku's cache minimum
PADDING=$(yes 'background context follows ' | head -1000 | tr -d '\n')
printf '{"model":"claude-haiku-4-5","max_tokens":32,"system":[{"type":"text","text":"%s","cache_control":{"type":"ephemeral"}}],"messages":[{"role":"user","content":"Say pong."}]}\n' "$PADDING" > /tmp/ct.json

# Call twice in a row, well within cache TTL
curl -sS -H "Authorization: Bearer $KEY" -H "anthropic-version: 2023-06-01" -H "content-type: application/json" -d @/tmp/ct.json http://localhost:2099/v1/messages
curl -sS -H "Authorization: Bearer $KEY" -H "anthropic-version: 2023-06-01" -H "content-type: application/json" -d @/tmp/ct.json http://localhost:2099/v1/messages
```

  **Expected:** Call 1 reports `cache_creation_input_tokens > 0`; Call 2 reports `cache_read_input_tokens > 0`.

  **Actual:** Both calls report `cache_creation_input_tokens: 0` and `cache_read_input_tokens: 0`. Verified both in the response usage object AND in the `agent_messages` table.

  ```sql
  SELECT timestamp, model, input_tokens, cache_read_tokens, cache_creation_tokens
  FROM agent_messages ORDER BY timestamp DESC LIMIT 2;

          timestamp        |           model           | input_tokens | cache_read_tokens | cache_creation_tokens
  -------------------------+---------------------------+--------------+-------------------+-----------------------
   2026-05-11 03:29:18.824 | claude-haiku-4-5-20251001 |         3013 |                 0 |                     0
   2026-05-11 03:28:43.434 | claude-haiku-4-5-20251001 |         3013 |                 0 |                     0
  ```

guillaumegay13:
### ✨ What changed
- `POST /v1/messages` responses now report real `cache_creation_input_tokens` and `cache_read_input_tokens`. They previously always read `0`.
- `agent_messages` rows store the same cache counts Anthropic returned, instead of `0` for both columns.
- Dedup no longer double-counts cache tokens when checking whether a success was already recorded.
- New e2e regression: real `POST /v1/messages` through the proxy stack against a stubbed Anthropic upstream.

### 💭 Why
`/v1/messages` is the Anthropic Messages API entry point added in #1810 so clients like Claude Code and `@anthropic-ai/sdk` can use Manifest via `ANTHROPIC_BASE_URL`. Internally the proxy translates the incoming request into chat-completions for routing, then translates the response back into Anthropic shape on the way out.

That back-translation was zeroing the cache creation counter, and the recorder was reading cache reads from the wrong field. So even when Anthropic actually served from its prompt cache, both the response and the dashboard reported `0/0`. The bug only affected `/v1/messages`; `/v1/chat/completions` and `/v1/responses` don't go through that conversion.

### 👤 For users
Anthropic-SDK clients pointing at Manifest see real cache token counts in responses. The Messages page in the dashboard shows accurate cache reads / cache creation per row.

### 🔧 For operators
No migration, no env vars. The dedup change may suppress duplicates that previously slipped through; nothing else moves.

### 📝 Notes
- Closes #1871.
- Cost calculation still prices cache reads at the normal input rate. Separate follow-up.
- 4459 unit + 149 e2e pass, 100% line coverage on changed source files, codex review clean.


guillaumegay13:
@brunobuddy ready to be merged

PR #1898, opened 2026-05-12
fix(routing): return 409 instead of 500 on concurrent custom-provider create

calebcgates:
## Summary

When two clients call `POST /api/v1/routing/:agentName/custom-providers` simultaneously with the same provider name, the second request currently returns **HTTP 500** because the unique-violation `QueryFailedError` from `repo.insert(cp)` bubbles past the existing pre-check guard. This PR wraps the insert and translates unique-violation errors into the same `ConflictException` (HTTP 409) that the pre-check already produces, so concurrent creates are indistinguishable from sequential ones from the client's perspective.

## Why

`CustomProvider` declares `@Index(['agent_id', 'name'], { unique: true })` (`packages/backend/src/entities/custom-provider.entity.ts:15`). `CustomProviderService.create()` does a find-then-insert pre-check: it calls `findOne({ where: { agent_id, name } })` and throws `ConflictException` if a row exists, then later runs `insert(cp)`. The pre-check is racy — between `findOne` returning null and `insert` running, a parallel request can land the same `(agent_id, name)`. The second insert hits the unique index, postgres raises SQLSTATE `23505`, typeorm wraps it as `QueryFailedError`, and Nest's default exception filter renders it as **500 Internal Server Error**.

The codebase already handles this exact race for `Agent` creation in `packages/backend/src/analytics/controllers/agents.controller.ts:75-78` and `:117-119` (`onboardAgent` and `duplicate` both wrap their service call in `try { ... } catch (e) { if (e instanceof QueryFailedError && /unique|duplicate/i.test(e.message)) throw new ConflictException(...); throw e; }`). The pattern is tested in `packages/backend/src/analytics/controllers/agents.controller.spec.ts:506-558`. This PR adopts the same convention for the symmetric race in `CustomProviderService.create()`.

## Solution

- Extend the existing `import { Repository } from 'typeorm'` to also import `QueryFailedError`.
- Wrap the `repo.insert(cp)` call in `try { ... } catch (error) { if (error instanceof QueryFailedError && /unique|duplicate/i.test(error.message)) { throw new ConflictException(...) } throw error; }`.
- Re-use the same message string the pre-check throws so client-side error handlers can dedupe.
- Two new unit tests in `custom-provider.service.spec.ts` — one positive (unique-violation → 409), one negative (non-unique `QueryFailedError` re-thrown unchanged so `connection refused` etc. still surface as 500 + observability events).

No other call-sites in `custom-provider.service.ts` are touched. No changes to `update()`, `remove()`, or any read paths.

## Testing

- `npm test --workspace=packages/backend` — all **4454+** tests still pass; two new tests added to the existing `describe('create', ...)` block (60/60 in the target spec).
- `npm run lint --workspace=packages/backend` — clean (no new warnings in the modified files).
- `npm run build --workspace=packages/backend` — `nest build` green; the import extension is type-safe.
- Changeset added under `.changeset/` targeting `manifest` (patch).

I did not run the e2e suite (`npm run test:e2e --workspace=packages/backend`) locally because it requires a running Postgres. The unit tests cover both branches of the new try/catch directly.

---
## Summary by cubic
Return 409 Conflict instead of 500 when two clients concurrently create the same custom provider via POST /api/v1/routing/:agentName/custom-providers. Unique-violation from insert is now translated to a ConflictException so concurrent and sequential creates behave the same.

- **Bug Fixes**
  - Wrap `repo.insert` and map duplicate/unique `QueryFailedError` to `ConflictException` with the same message as the pre-check; non-unique errors still bubble up.
  - Added tests for the conflict path and non-unique pass-through, matching the pattern used in agent creation.

<sup>Written for commit 10a5858f1feee2ee002bfc55cb6ef6937a1c62fd. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/custom-provider-409-on-concurrent-create.md
+++ b/.changeset/custom-provider-409-on-concurrent-create.md
@@ -1,0 +1,5 @@
+---
+'manifest': patch
+---
+
+Translate concurrent-insert unique-violation on `POST /api/v1/routing/:agentName/custom-providers` into HTTP 409 instead of 500. The find-then-insert pre-check in `CustomProviderService.create()` was racy against the `(agent_id, name)` unique index — two parallel requests with the same provider name could both pass the pre-check, then the second insert raised `QueryFailedError` and bubbled past the controller as a 500. Now the insert is wrapped and unique-violation `QueryFailedError`s are translated into the same `ConflictException` the pre-check produces, matching the convention already used for the symmetric race on `Agent` creation in `agents.controller.ts`. Non-unique errors (connection drops, etc.) continue to propagate unchanged.
--- a/packages/backend/src/routing/custom-provider/custom-provider.service.spec.ts
+++ b/packages/backend/src/routing/custom-provider/custom-provider.service.spec.ts
@@ -6,7 +6,7 @@ jest.mock('../../common/utils/detect-self-hosted', () => ({
 }));
 
 import { BadRequestException, ConflictException, NotFoundException } from '@nestjs/common';
-import { Repository } from 'typeorm';
+import { QueryFailedError, Repository } from 'typeorm';
 import { CustomProviderService } from './custom-provider.service';
 import { CustomProvider } from '../../entities/custom-provider.entity';
 import { ProviderService } from '../routing-core/provider.service';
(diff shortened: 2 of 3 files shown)


guillaumegay13:
hey @calebcgates thank you for contributing! which issue does it solve?
and I'm curious : how do you create 2 concurrent providers simultaneously ?

calebcgates:
> hey @calebcgates thank you for contributing! which issue does it solve? and I'm curious : how do you create 2 concurrent providers simultaneously ?

Hey @guillaumegay13, thanks for the quick look!

**Re: the issue.** No specific tracked issue — I found this by spotting an asymmetry in the codebase: `agents.controller.ts:75-78,117-119` already translates `QueryFailedError(unique|duplicate)` → `ConflictException` for `Agent` creation/duplication (tested in `agents.controller.spec.ts:506-558`), but `CustomProviderService.create()` doesn't do the same for the symmetric race against `@Index(['agent_id', 'name'], { unique: true })`. The PR closes that gap. Happy to open a tracking issue retroactively if you'd prefer.

**Re: how the race triggers in practice.** The window is small — the ms between `findOne` returning null and `insert` landing — but reachable through normal UI patterns, not just intentional concurrent calls:

- **Double-click on "Save"** in the create-provider form (most common; happens whenever the button isn't disabled mid-request)
- **Slow-network retry** — user hits Save, browser stalls, user clicks again before the first reply arrives
- **Scripted setup** — CLI / Terraform-style tools that create providers in parallel
- **Multi-tab edits** of the same form

When it lands, the client currently sees 500 — which it treats as "server broke, retry later" — instead of 409, which the client already knows how to handle from the pre-check path. The two new unit tests in `custom-provider.service.spec.ts` reproduce the failure directly by mocking `repo.insert()` to reject with `QueryFailedError(unique constraint)`, which is exactly what the DB raises in production. Happy to add a controller-level integration test (parallel `POST`s through a `TestingModule`) if you'd find that more convincing.

PR #1900, opened 2026-05-13
Add Codex CLI as a coding-assistant agent platform

kyya:
> **Draft** — opening this early to get maintainer guidance on contribution requirements (changeset, Wingman profile, coverage gap, scope split) before polishing.

## Problem

OpenAI's [Codex CLI](https://github.com/openai/codex) is a Responses-API-only client built in Rust. Pointing it at a self-hosted Manifest instance currently fails with:

```
ERROR: stream disconnected before completion: stream closed before response.completed
```

Codex retries 5 times, then gives up. Non-streaming `POST /v1/responses` works fine; the breakage is in the streaming path. Codex's SSE parser strictly requires the full OpenAI Responses lifecycle envelope — `response.created` → `output_item.added` → `output_text.delta(*)` → `output_item.done` → `response.completed` with a populated `output[]` — and the current chat→responses encoder in `responses-adapter.ts` only emits `output_text.delta` and `response.completed` with `output: []`. That is enough for the OpenAI SDK and works for `/v1/messages` (Claude Code), but Codex tears the stream down.

Two more upstream-compat issues surface once Codex is reaching a non-OpenAI backend through Manifest:

1. Codex always wires its instructions as `role: "developer"`. DeepSeek, MiniMax, Z.AI, and most other OpenAI-compat chat/completions upstreams still only accept `{system,user,assistant,tool}` and 400 on anything else.
2. Codex advertises OpenAI-hosted tools (`web_search`, `file_search`, `computer_use_preview`, ...) every turn. Chat/completions upstreams reject any tool whose `type` isn't `function`.

## Approach

Add `codex` as a new `coding`-category agent platform in `packages/shared/src/agent-type.ts`. When the calling agent's `agent_platform === 'codex'`, the proxy routes streaming `/v1/responses` through a **new stateful transformer** (`createStrictChatToResponsesTransformer`) that emits the full lifecycle. Other platforms keep the existing delta-only encoder unchanged — zero behaviour change for OpenAI SDK, OpenClaw, etc.

`agent_platform` is propagated end-to-end: `IngestionContext` gains an `agentPlatform` field, `AgentKeyAuthGuard` selects + caches `agent.agent_platform`, the controller forwards it to `handleStreamResponse`, which picks the encoder.

The role normalization and hosted-tool filter are applied unconditionally in `toChatCompletionsRequest` — they fix genuine OpenAI-compat bugs that any caller using developer messages or hosted tools would hit, not just Codex.

## Commits

| Commit | What |
|---|---|
| [`a366ac0`](https://github.com/silicondawn/manifest/commit/a366ac0a1) | feat: add Codex CLI agent platform + strict Responses-API SSE lifecycle |
| [`eaaa0b1`](https://github.com/silicondawn/manifest/commit/eaaa0b19d) | fix: translate `role: developer` → `system`, drop hosted tools |
| [`b2fb252`](https://github.com/silicondawn/manifest/commit/b2fb25206) | chore: switch Codex icon to lobehub mark |
| [`d60e4e3`](https://github.com/silicondawn/manifest/commit/d60e4e338) | chore: switch to lobehub Codex color variant |

(The two icon commits can be squashed if preferred — left split for diff clarity.)

## Tests

- **New strict-transformer fixtures** (5): full lifecycle ordering, empty-stream finalize, stable response/message ids, double-finalize idempotency, upstream model propagation
- **Updated** existing `toChatCompletionsRequest` fixture — hosted `web_search_preview` is now dropped
- **New** developer→system role normalization fixture
- **New** hosted-tools filter fixture
- **Updated** `IngestionContext` assertion in `agent-key-auth.guard.spec.ts` to include `agentPlatform`
- **New** codex platform propagation test in the guard
- **New** `handleStreamResponse` branch tests covering codex vs non-codex
- **Updated** `agent-type.spec.ts` to assert codex is in the coding category with a labelled icon

Local results: `npm test --workspace=packages/backend` → **4543/4543 passing**; `npm test --workspace=packages/shared` → **22/22 passing**; `turbo build` → 3/3 packages green.

## End-to-end verification

1. Local dev server (`packages/backend/dist/main.js` against a fresh `manifest_codextest` Postgres):
   - Created a coding/codex agent via `POST /api/v1/agents`, attached a real DeepSeek API key
   - Codex CLI 0.130 with `base_url=http://127.0.0.1:3001/v1` and the agent's `mnfst_*` token completed `"say hello in 3 words"` cleanly, no retries
2. Live rollout on my own self-hosted instance (built from this branch):
   - Pulled, switched live container, healthy, Codex hello succeeded against the new image

## What I deliberately left for maintainer guidance

Opening as a **draft** because there are a few project-specific things I haven't done and want to defer to the maintainers' preference rather than guess:

1. **No `changeset` entry yet.** Should I add one targeting `manifest` (and at what bump level)? Codex support is user-facing and additive, so my instinct is `minor`, but happy to be told.
2. **No Wingman profile for Codex.** `packages/wingman/` ships 7 profiles modelled on real client headers/body shapes. Codex is now first-class on the routing side; would you like a Codex profile added in this PR, a follow-up, or not at all (Codex is a Rust binary, not really useful to replay from a TS playground)?
3. **Coverage.** I added tests for the new code paths but didn't run `--coverage` end-to-end against the new branches yet. If CI flags uncovered lines, point me at the gap and I'll cover them.
4. **Dashboard help / setup-modal copy.** I added the platform to `AGENT_PLATFORMS` so it shows up in `AgentTypeGrid` automatically (data-driven), but I didn't add a setup-modal snippet for "how to point Codex CLI at Manifest". Want me to add that here or in a follow-up doc PR?
5. **Scope.** If splitting helps review, I'd suggest:
   - PR1: `feat:` Codex platform + strict SSE encoder
   - PR2: `fix:` developer→system + hosted-tool filter (genuinely useful independent of Codex)
   - PR3: icon
   
   Happy to do whichever you prefer.

## Closes / Refs

Not aware of an existing issue tracking Codex support; happy to file one and link if useful.

---
## Summary by cubic
Adds `codex` as a first-class coding agent and streams the full OpenAI Responses SSE lifecycle (including function_call) so Codex CLI works end-to-end. Also adds a setup panel with a copyable `~/.codex/config.toml` snippet and refines tool handling to preserve hosted tools on native Responses paths.

- **New Features**
  - Registered `codex` with label/icon and a setup panel with a copyable config (`wire_api = "responses"`, `OPENAI_API_KEY`).
  - For streaming `/v1/responses`, when `agentPlatform === 'codex'` use a strict, stateful transformer that emits the full lifecycle: response.created → in_progress → output_item/content_part added → output_text and function_call argument deltas → *_done → response.completed with populated `output[]` and usage. Non-Codex stays delta-only. Propagate `agentPlatform` via `IngestionContext`, cache it in the auth guard, and pass it to `handleStreamResponse`.

- **Bug Fixes**
  - Normalize Responses `role: developer` to chat/completions `system`.
  - Move hosted tool filtering to the chat-completions sanitizer: chat-completions forwards only `function` tools; native Responses paths preserve hosted tools (`web_search`, `file_search`, etc.).
  - Align SSE shape with OpenAI Responses: stamp `response_id` on all per-item events and a monotonic `sequence_number` on every event; include both `arguments` and a final `item` on `response.function_call_arguments.done` (with top-level `name`); mark function_call `output_item.added` as `status: "in_progress"` and flip to `completed` on `.done`; replay buffered argument fragments when headers arrive late.

<sup>Written for commit e818f7604849305cf92743adf26d99710e2aece0. Summary will update on new commits. <a href="https://cubic.dev/pr/mnfst/manifest/pull/1900?utm_source=github">Review in cubic</a></sup>


guillaumegay13:
hi @kyya thanks for the PR - Codex support as Coding assistant is definitely a great addition for Manifest!

Did you test your PR with Codex tools calls, not only simple text prompts ?

kyya:
Thanks for catching this — you were right to ask. Running it against codex 0.130.0 with real tool calls, I realized the SSE transformer was dropping `delta.tool_calls` (only text deltas were being forwarded), so codex saw no function_calls and ended the turn. Pushed `3bbe56dff` to handle the tool_call lifecycle properly; it now dispatches multi-turn shell tools end-to-end. Spec covers the obvious cases but happy to add more if you spot any I missed.

Also threw in `0e4306af4` for the empty Codex panel in the setup modal (`SetupStepAddProvider` had no Match case for `codex`). Happy to split that off into its own PR if you'd rather keep this focused.

guillaumegay13:
Thanks, this is looking much better after the tool-call fix.

One small protocol hardening request before merge: could you align the synthetic function-call SSE events a bit closer to the Responses stream shape?

- Include `response_id` on `response.output_item.added`, `response.function_call_arguments.delta`, and `response.function_call_arguments.done`.
- Make `response.function_call_arguments.done` carry the final function-call `item`, not only top-level `arguments`.
- When args arrive before both id/name are known, replay the buffered argument fragments once `output_item.added` is emitted, matching the comment.

Codex appears lenient today because it mainly consumes `output_item.done`, so this is not a big design objection, but Manifest should avoid emitting a near-Responses stream with small shape mismatches.

kyya:
@guillaumegay13 Thanks for the protocol-fidelity catch — all three are addressed in 1f35e86.

1. `response_id` is now stamped on every per-item event: `output_item.added`, `content_part.added`/`done`, `output_text.delta`/`done`, `output_item.done`, and both `function_call_arguments.delta` / `.done`. The lifecycle envelopes (`response.created` / `in_progress` / `completed`) keep the id inside `response.id` to match OpenAI's stream.

2. `response.function_call_arguments.done` now carries both the top-level `arguments` string and a final structured `item`, so SDK consumers that prefer the structured form get the call without waiting for `output_item.done` or `response.completed`.

3. Argument fragments arriving before `id` + `name` are now replayed as a single coalesced `function_call_arguments.delta` the moment `output_item.added` fires — matching the in-code comment that already promised this behavior. Without the replay, the opening fragment was only observable in `response.completed`.

Verification:
- Unit: 3 new cases in `responses-adapter.spec.ts` pin the new shape (response_id on all relevant events, `done` carrying `item`, args replay after deferred `added`).
- Runtime: hit `/v1/responses` against a codex-platform agent with curl and via codex 0.130.0 CLI; first-turn tool dispatch reaches the shell tool intact and all SSE events carry response_id.

(Tangent noticed while verifying: the strict chat→responses transformer also drops `delta.reasoning_content`, so DeepSeek thinking-mode second turns 400 — same class of bug being addressed for the Anthropic Messages path in #1907; the codex/Responses path will need the matching treatment in a follow-up.)

kyya:
Follow-up in 0923331 — self-audited the synthetic stream against `openai-python`'s `ResponseStreamEvent` types and closed three more required-field gaps that strict pydantic consumers would have caught:

1. `sequence_number` (monotonic, zero-indexed) on every emitted event. Declared `required` on every `Response*Event` variant in `openai-python/src/openai/types/responses/`; strict consumers reject events that omit it. Implemented as a closure-scoped `emit()` helper so no call site can skip or double-stamp.

2. Top-level `name` on `response.function_call_arguments.done`. Required alongside `arguments` and `item_id` by `ResponseFunctionCallArgumentsDoneEvent`.

3. `status: "in_progress"` on the `function_call` item payload inside `response.output_item.added`. Real Responses streams open the item `in_progress` and flip it to `completed` only on `output_item.done` — we previously omitted the field on `added`.

Three new spec cases pin the new shape (`responses-adapter.spec.ts`: 42/42, full `routing/proxy` suite: 1315/1315). Re-ran the curl + codex 0.130.0 CLI smoke against a freshly built image — first-turn tool dispatch still intact, all events now carry `sequence_number` and the new fields.

guillaumegay13:
@kyya quick question : which models would you like to use Codex with ?
As it's dropping OpenAI native tools like web_search, is it really relevant to use other OpenAI models ? I assume it'll drastically drop the output quality.

kyya:
@guillaumegay13 The strip is narrower than it looks — Codex requests routed to a Responses-API-native upstream (`openai-responses`, `openai-subscription`, `copilot-responses`) take the `chatgpt` format branch in `provider-client.buildRequest()`, use the original body via `toNativeResponsesRequest`, and keep hosted tools intact. The non-`function` filter only fires on the chat-completions path, and chat-completions' own `tools[]` schema rejects non-function entries (OpenAI's own `/v1/chat/completions` included), so the strip is per-protocol, not per-vendor.

Took the opportunity to clean up the layering in e818f76 — moved the hosted-tool filter out of `toChatTools` (now schema-rewrite only) and into `sanitizeOpenAiBody`, which is the per-provider chat-completions sanitizer that already owns `_anthropic*` stripping, `max_tokens` normalization, `reasoning_content` preservation, etc. Behavior-equivalent for every existing path (the upstream body is identical), but the responsibility now lives where it belongs.

The useful follow-up for actually preserving hosted tools across more upstreams isn't "is the upstream OpenAI" — it's "does the upstream speak Responses-API natively". OpenRouter's `/api/v1/responses` (currently beta, with explicit hosted-tool support including `web_search`) and self-hosted Responses-compatible gateways would all preserve hosted tools if we let providers declare that capability via a `support_responses_api` flag on `user_providers` / `custom_providers` — routing would then send the request down the `toNativeResponsesRequest` path instead of the sanitizer. Worth a separate PR with a concrete user; this one's scope stays the non-OpenAI chat-completions path.

luandro:
Any updates on this? Would really be a great addition.

guillaumegay13:
hi @luandro 

thanks for reaching out

I didn't get the time to properly test it. I'll have a look and let you know!

Issue #1899, opened 2026-05-12
OpenCode Go: preserve reasoning_content for non-DeepSeek reasoning models


xtraxion:
## Bug Description

Manifest 6.3.0 preserves `reasoning_content` for DeepSeek models routed through OpenCode Go, but still strips it for other OpenCode Go reasoning model families.

This means plain chat can work, but multi-turn tool-use with thinking/reasoning can still fail or trigger fallback for non-DeepSeek OpenCode Go models such as Kimi, GLM, Qwen, MiMo and MiniMax.

Related:
- #1862 fixed the DeepSeek/OpenCode Go case
- #1872 preserves DeepSeek `reasoning_content` through aggregator routes
- #1811 appears to be a broader open PR around cache/reinject, but 6.3.0 still only checks for `deepseek` in the current released code path

## Current behavior in 6.3.0

The released `supportsReasoningContent()` logic preserves `reasoning_content` only when the bare model slug contains `deepseek`:

```ts
const bare = model.toLowerCase().split('/').pop() ?? '';
return bare.includes('deepseek');
```

So this works:

- `opencode-go/deepseek-v4-pro` ✅

But these are still stripped:

- `opencode-go/kimi-k2.6` ❌
- `opencode-go/glm-5.1` ❌
- `opencode-go/qwen3.6-plus` ❌
- `opencode-go/minimax-m2.7` ❌
- `opencode-go/mimo-v2.5` ❌

## Why this matters

The issue does not necessarily show up in a simple chat test. It appears when both are true:

1. thinking/reasoning mode is enabled
2. the conversation includes tool calls

When an assistant message contains both `reasoning_content` and `tool_calls`, the follow-up request after the tool result needs to replay that assistant message with `reasoning_content` intact. If Manifest strips it, upstream providers can reject the request and Manifest falls back to another route/model.

## Reproduction / sanitizer test

Tested against Manifest `manifest@6.3.0`:

```js
const { sanitizeOpenAiBody } = require('./dist/routing/proxy/provider-client-converters');

const body = {
  messages: [
    { role: 'user', content: 'x' },
    {
      role: 'assistant',
      content: '',
      reasoning_content: 'thinking',
      tool_calls: [
        {
          id: 'call_1',
          type: 'function',
          function: { name: 'foo', arguments: '{}' },
        },
      ],
    },
    { role: 'tool', tool_call_id: 'call_1', content: '{}' },
  ],
};

for (const model of [
[...]

Issue #1862, opened 2026-05-08
[BUG] opencode-go/deepseek-v4-pro "reasoning content" change needed


FirebirdRender:
failing 100% at the moment.

```
{"error":{"message":"Error from provider (DeepSeek): The `reasoning_content` in the thinking mode must be passed back to the API.","type":"invalid_request_error","param":null,"code":"invalid_request_error"}}
Fallback from custom:f0c38cb6-34e9-40d8-b639-4734195d2e82/mistralai/mistral-large-3-675b-instruct-2512 (attempt #1)
Provider: OpenCode Go
Auth: subscription
Model: Deepseek V4 Pro
Model ID: opencode-go/deepseek-v4-pro
Routing: reasoning
Reason: formal_logic_override

Request Headers
Header	Value
accept	*/*
accept-encoding	gzip, deflate, br, zstd
connection	keep-alive
content-length	184015
content-type	application/json
user-agent	opencode/1.4.7 ai-sdk/provider-utils/4.0.23 runtime/bun/1.3.11
```

xtraxion:
exact same for Kimi K2.6
{"error":{"message":"Error from provider: Provider returned error","code":400,"metadata":{"raw":"{\"error\":{\"message\":\"thinking is enabled but reasoning_content is missing in assistant tool call message at index 2\",\"type\":\"invalid_request_error\"}}","provider_name":"Moonshot AI","is_byok":true}},"user_id":"user_2z4xm5LomaIHfsnVqMhFsWrVrGY"}

Issue #992, opened 2026-03-05
bug: DeepSeek reasoning models return 400 missing reasoning_content field


guillaumegay13:
When the proxy routes to DeepSeek R1, multi-turn conversations with tool calls fail with: 400 Missing reasoning_content field in the assistant message.

Ref: https://api-docs.deepseek.com/guides/thinking_mode#tool-calls

DeepSeek thinking mode requires every assistant message to include reasoning_content (even if empty). Clients don't include this field, and
the proxy passes messages through as-is.

guillaumegay13:
@brunobuddy I think you fixed it - please confirm and feel free to close the issue

brunobuddy:
@guillaumegay13 yes fixed in v5.25.3

PR #1907, opened 2026-05-13
fix(proxy): emit DeepSeek reasoning_content as Anthropic thinking blocks in stream

kyya:
## Summary

When a client hits the Anthropic Messages endpoint (`POST /v1/messages`) and the request is routed to a DeepSeek (or DeepSeek-family) model in thinking mode, the **streaming** code path silently drops every `delta.reasoning_content` chunk. Clients never see a `thinking` content block, so they can't echo the trace back on the next turn — and the next request comes back as:

```
400 invalid_request_error
The `reasoning_content` in the thinking mode must be passed back to the API.
```

The non-streaming response path already translates `message.reasoning_content` into a leading `{type: "thinking", thinking: "..."}` content block (see `chatCompletionsResponseToMessages` in `anthropic-messages-adapter.ts`). The stream transformer in the same file just didn't have the corresponding logic — `transformStreamChunk` only branched on `delta.content` and `delta.tool_calls`. With Claude CLI defaulting to `stream: true`, every thinking-mode multi-turn conversation that flowed through `/v1/messages` was broken.

Repro in production traffic:

```
ProxyResponseHandler] Upstream error 400: provider=deepseek model=deepseek-v4-flash tier=default
body={"error":{"message":"The `reasoning_content` in the thinking mode must be passed back to the API.",
"type":"invalid_request_error","param":null,"code":"invalid_request_error"}}
HttpErrors] 400 POST /v1/messages?beta=true 514ms ua=claude-cli/2.1.140
```

## What this PR does

In `createMessagesStreamTransformer` (`packages/backend/src/routing/proxy/anthropic-messages-adapter.ts`):

- Track a `thinkingIndex` / `thinkingOpened` on `StreamState`.
- When a chunk carries `delta.reasoning_content`, emit Anthropic-native stream events:
  - `content_block_start` with `{type: "thinking", thinking: ""}`
  - `content_block_delta` with `{type: "thinking_delta", thinking: "..."}`
- Close the thinking block via `content_block_stop` **before** opening either a `text` or `tool_use` block, so the stream never has two open content blocks at once (Anthropic spec).
- Close any still-open thinking block in the final `closeStream` path (e.g. responses that contain only reasoning and never produce a content payload).
- Adjust `nextBlockIndex` so the thinking block reserves index 0 when present.

The shape mirrors what `chatCompletionsResponseToMessages` already produces for non-streaming responses, so streaming and non-streaming paths now agree.

## Why thinking *before* text, not after

DeepSeek streams `delta.reasoning_content` first and `delta.content` second. Surfacing reasoning as a leading block (index 0) preserves that order in the Anthropic event stream, matches the non-streaming projection in this same file, and lines up with how the Anthropic SDK's `thinking` blocks are placed.

## Tests

`packages/backend/src/routing/proxy/__tests__/anthropic-messages-adapter.spec.ts` gets three new cases under `createMessagesStreamTransformer`:

1. `translates DeepSeek-style delta.reasoning_content into a leading thinking block` — full `message_start → thinking_start → thinking_delta × 2 → thinking_stop → text_start → text_delta → text_stop → message_delta → message_stop` ordering.
2. `closes the thinking block on finalize when the stream emits only reasoning_content` — reasoning-only response still emits a clean `content_block_stop`.
3. `closes the thinking block before opening a tool_use block` — when the model goes from reasoning straight into a tool call, thinking@0 is stopped before tool_use@1 starts.

All 36 tests in the suite pass. Line coverage on `anthropic-messages-adapter.ts` stays at 100%.

## Related

- #992 — same upstream error message but on the OpenAI chat_completions path (clients omitting `reasoning_content` on history).
- #1862 / #1899 / PR #1872 — preserve `reasoning_content` through aggregator routes (sanitization, not stream translation). The stream-translation gap fixed here is independent of those — `reasoning_content` is allowed through to DeepSeek correctly today; the issue is that the client never received the trace in the first place because the streaming Anthropic adapter dropped it.

## Test plan

- [x] `npx jest packages/backend/src/routing/proxy/__tests__/anthropic-messages-adapter.spec.ts` — 36/36 pass.
- [x] Local repro script feeds a synthetic DeepSeek SSE stream through the transformer and prints the translated Anthropic events; ordering and block-stop semantics verified.
- [x] Deployed to a self-hosted instance against real DeepSeek traffic from Claude CLI; the `400 The reasoning_content in the thinking mode must be passed back to the API` errors stopped reproducing after the upgrade.

---
## Summary by cubic
Fixes the Anthropic Messages streaming adapter to emit DeepSeek `delta.reasoning_content` as `thinking` blocks, handle interleaved chunks with fresh monotonic indices, and drop late reasoning during an open `tool_use`. This stops 400s in thinking mode and aligns streaming with non‑streaming behavior.

- **Bug Fixes**
  - In `packages/backend/src/routing/proxy/anthropic-messages-adapter.ts`, translate `delta.reasoning_content` into `thinking` blocks (start → `thinking_delta` → stop) and close it before `text`/`tool_use`.
  - Enforce one open block at a time; on interleaving, close and reopen `thinking`/`text` with a stream‑wide monotonic index counter.
  - Drop `reasoning_content` that arrives while any `tool_use` is open to keep the tool call contiguous and replayable; fragmented args for one tool call stay on a single `tool_use` block.
  - Close any open `thinking`/`text`/`tool_use` blocks on finalize; add tests for ordering, reasoning‑only finalize, interleaved reopen (both directions), dropped late reasoning during `tool_use`, and monotonic indices.

<sup>Written for commit d0ef03caa5ad04a8aea0686921418cf8a45c2df8. Summary will update on new commits. <a href="https://cubic.dev/pr/mnfst/manifest/pull/1907?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/packages/backend/src/routing/proxy/__tests__/anthropic-messages-adapter.spec.ts
+++ b/packages/backend/src/routing/proxy/__tests__/anthropic-messages-adapter.spec.ts
@@ -786,13 +786,14 @@ describe('Anthropic Messages adapter', () => {
       expect(events[9].data).toEqual({ type: 'content_block_stop', index: 2 });
     });
 
-    it('defers reasoning_content arriving during a tool_use and flushes it as a thinking block after the tool_use closes', () => {
-      // After a tool_use block has been started, late reasoning_content chunks
-      // are *buffered* rather than splitting the tool_use across two Anthropic
-      // blocks (each Anthropic tool_use block must carry its own complete
-      // input JSON — splitting would break client-side reassembly). The
-      // accumulated reasoning is emitted as a self-contained thinking block
-      // once the tool_use closes (here, on finalize via finish_reason).
+    it('drops reasoning_content arriving during an open tool_use to keep the tool_use block contiguous', () => {
+      // Manifest's Anthropic-compat layer assumes thinking blocks precede
+      // tool_use and replays them in that shape on the next turn. Emitting a
+      // post-tool thinking block would give clients a transcript shape
+      // (`thinking → tool_use → thinking`) we cannot safely replay, so late
+      // reasoning_content arriving while a tool_use is in-progress is dropped
+      // silently. The pre-tool thinking block and the tool_use itself stay
+      // intact.
       const t = createMessagesStreamTransformer('deepseek-v4-flash');
       const sse = flushChunks(t, [
         'data: {"choices":[{"delta":{"reasoning_content":"plan A"}}]}\n\n',
@@ -819,32 +820,32 @@ describe('Anthropic Messages adapter', () => {
         'content_block_delta', // thinking_delta "plan A"
         'content_block_stop', // thinking@0 closed by tool_use
         'content_block_start', // tool_use@1
-        'content_block_delta', // input_json_delta
+        'content_block_delta', // input_json_delta — "plan B" / " continued" dropped while open
         'content_block_stop', // tool_use@1 closed on finalize
-        'content_block_start', // thinking@2 — deferred reasoning flush
-        'content_block_delta', // single combined "plan B continued" delta
-        'content_block_stop', // thinking@2 closed
         'message_delta',
         'message_stop',
       ]);
       expect(events[4].data).toMatchObject({ index: 1, content_block: { type: 'tool_use' } });
       expect(events[6].data).toEqual({ type: 'content_block_stop', index: 1 });
-      expect(events[7].data).toMatchObject({ index: 2, content_block: { type: 'thinking' } });
-      expect(events[8].data).toMatchObject({
-        index: 2,
-        delta: { type: 'thinking_delta', thinking: 'plan B continued' },
-      });
-      expect(events[9].data).toEqual({ type: 'content_block_stop', index: 2 });
+      // No thinking block ever opens after tool_use@1.
+      const postToolThinking = events.find(
+        (e, i) =>
+          e.event === 'content_block_start' &&
+          i > 6 &&
+          (e.data.content_block as { type?: string })?.type === 'thinking',
+      );
+      expect(postToolThinking).toBeUndefined();
     });
 
-    it('keeps a fragmented tool_call on a single tool_use block when reasoning_content interleaves between arg chunks', () => {
-      // Regression for #1907 review: with the previous logic, late
-      // reasoning_content closed the in-progress tool_use, then the next
-      // tool_call delta for the same `index` re-emitted `content_block_start`
-      // against the same (already stopped) tool_use index, breaking
-      // monotonicity. We now buffer the reasoning so the tool_use stays open
-      // and accumulates all its arg chunks into a single Anthropic block —
-      // the only shape where the client can reassemble the input JSON.
+    it('keeps a fragmented tool_call on a single tool_use block while dropping interleaved reasoning_content', () => {
+      // Regression for the #1907 review: late reasoning_content used to close
+      // the in-progress tool_use, then the next arg chunk for the same `index`
+      // re-emitted `content_block_start` against the already-stopped index,
+      // breaking monotonicity. Per the reviewer's follow-up, we now keep the
+      // tool_use contiguous AND drop the interleaved reasoning entirely
+      // rather than flushing a post-tool thinking block — emitting
+      // `thinking → tool_use → thinking` would produce an Anthropic-shaped
+      // transcript we cannot safely replay on the next turn.
       const t = createMessagesStreamTransformer('deepseek-v4-flash');
       const sse = flushChunks(t, [
         'data: {"choices":[{"delta":{"reasoning_content":"plan"}}]}\n\n',
@@ -874,9 +875,6 @@ describe('Anthropic Messages adapter', () => {
         'content_block_delta', // input_json_delta "{\"q\":"
         'content_block_delta', // input_json_delta "1}" — same block, no reopen at @1
         'content_block_stop', // tool_use@1 stopped on finalize
-        'content_block_start', // thinking@2 — deferred "late"
-        'content_block_delta', // thinking_delta "late"
-        'content_block_stop', // thinking@2 stopped
         'message_delta',
         'message_stop',
       ]);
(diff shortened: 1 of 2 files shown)


guillaumegay13:
thanks @kyya. Please fix latest cubic issue

kyya:
### Integration test — local Docker, real DeepSeek traffic

Built the image off this branch (`843c16bc0`), brought up `docker/docker-compose.yml` as-is, and ran the full path end-to-end against `api.deepseek.com`.

**Setup**
- Image: `manifestdotbuild/manifest:integration-1907` (distroless, 367 MB), tagged `:latest` for compose
- Stack: `mnfst-postgres-1` + `mnfst-manifest-1`, both healthy on `127.0.0.1:2099`
- Admin onboarded via `POST /api/v1/setup/admin`
- Agent `deepseek-test` created; DeepSeek connected with a live API key
- Tier auto-assign picked `deepseek-v4-pro` for `complex` / `reasoning`, `deepseek-v4-flash` for the rest

**Request** — `POST /v1/messages` with `stream: true`, `thinking: { type: "enabled", budget_tokens: 1024 }`, a math-induction prompt that forces reasoning.

**SSE event histogram**

```
   1  event: message_start
   2  event: content_block_start
 871  event: content_block_delta
   2  event: content_block_stop
   1  event: message_delta
   1  event: message_stop
```

**Block layout (the bit this PR actually fixes)**

```
content_block_start  index=0  {type:"thinking"}      ← was previously dropped
content_block_delta  index=0  {type:"thinking_delta", ...}  ×N
content_block_stop   index=0
content_block_start  index=1  {type:"text"}
content_block_delta  index=1  {type:"text_delta", ...}  ×M
content_block_stop   index=1
message_delta        stop_reason=end_turn, output_tokens=872
message_stop
```

Checks:
- thinking block opens at index 0 with the right `content_block` shape
- thinking is stopped before text starts (Anthropic SSE: at most one open block on the wire)
- all 871 `*_delta` events fall between their matching `start` / `stop` — no deltas against a stopped block, which is the failure mode cubic flagged on the late/interleaved path
- usage accounting flows back through (`input_tokens=49`, `output_tokens=872`)

guillaumegay13 (commented):
I think this still has one interleaving hole before merge.

The new monotonic block-index logic works for `thinking -> text -> thinking` and `thinking -> tool_use -> thinking`, but a resumed tool call can still reuse a stopped block index. Repro stream:

```ts
const t = createMessagesStreamTransformer('deepseek-v4-flash');
const chunks = [
  'data: {"choices":[{"delta":{"reasoning_content":"plan"}}]}\n\n',
  'data: {"choices":[{"delta":{"tool_calls":[{"index":0,"id":"tc_1","function":{"name":"search","arguments":"{\\\"q\\\":"}}]}}]}\n\n',
  'data: {"choices":[{"delta":{"reasoning_content":"late"}}]}\n\n',
  'data: {"choices":[{"delta":{"tool_calls":[{"index":0,"function":{"arguments":"1}"}}]}}]}\n\n',
  'data: {"choices":[{"delta":{},"finish_reason":"tool_calls"}]}\n\n',
];
```

On this PR head I get:

```text
thinking@0 start/delta/stop
 tool_use@1 start/delta/stop
 thinking@2 start/delta/stop
 tool_use@1 start/delta/stop  // reopened with old index
```

That reopens `tool_use` at `index:1` after a later `thinking@2`, so the stream is no longer monotonic and still violates the Anthropic content block ordering rules. The issue is that existing `state.toolCalls.get(callIndex)` entries keep their old `entry.index` after `closeOpenToolCalls()` sets `opened = false`, so later argument chunks for the same OpenAI tool call do not allocate a fresh content block index.

Can we add a regression for this case and either allocate a fresh Anthropic block index when reopening a stopped tool call, or avoid closing an in-progress tool call for late reasoning chunks?

kyya:
Pushed `49b7b6ca1`.

Went with option (b): we buffer `reasoning_content` while a `tool_use` is open instead of closing it. Splitting one OpenAI tool_call across two Anthropic `tool_use` blocks would leave each block with partial input JSON, so the client side can't reassemble it. With the buffer, the tool_use stays at a single block index across all arg chunks, and the deferred reasoning is flushed as a self-contained `thinking` block at a fresh monotonic index when the tool_use closes naturally (next text content or finalize).

Added a regression test mirroring your exact repro (`keeps a fragmented tool_call on a single tool_use block when reasoning_content interleaves between arg chunks`) — asserts the `content_block_start` index sequence is strictly `[0, 1, 2]` and `tool_use@1` accumulates both arg chunks. Updated the earlier "thinking after tool_use" test to reflect the new defer-then-flush ordering.

kyya:
Also ran the local integration test against a freshly built image of `49b7b6ca1` (a real DeepSeek-routed `POST /v1/messages` with `thinking: enabled`). Event histogram on the SSE output:

```
1 message_start
2 content_block_start   (thinking@0, text@1)
1023 content_block_delta
2 content_block_stop    (thinking@0, text@1)
1 message_delta
1 message_stop
```

Block indices stay strictly monotonic, thinking@0 stops before text@1 starts, and the `thinking_delta` stream is intact end-to-end on the happy path.

The specific fragmented-tool_call repro is hard to trigger in real traffic — it needs the model to interleave a `reasoning_content` chunk between two arg chunks of the *same* tool call, which we don't reliably see in the wild. The synthetic SSE in `keeps a fragmented tool_call on a single tool_use block when reasoning_content interleaves between arg chunks` is exactly your repro stream, so that's where it's locked in.

guillaumegay13:
thanks @kyya, I think the core fix is right: the normal DeepSeek streaming path was dropping `delta.reasoning_content`, so emitting the leading `thinking` block before text / tool use fixes a real bug.

One thing I’d change before merging: I don’t think we should emit a new `thinking` block after a `tool_use` when `reasoning_content` arrives while a tool call is already open. That gives clients a transcript like `thinking@0 -> tool_use@1 -> thinking@2` with `stop_reason=tool_use`, but our Anthropic compatibility code assumes thinking / redacted_thinking blocks precede `tool_use` and get replayed in that shape on the next turn.

I’d rather keep the `tool_use` block contiguous and drop / ignore that late reasoning fragment in this narrow edge case (or keep it out-of-band), instead of emitting an Anthropic-shaped stream that we probably can’t safely replay. The main fix should stay; I just think the deferred post-tool thinking flush is the wrong tradeoff.

kyya:
@guillaumegay13 Agreed — replay safety wins over surfacing the fragment. Switched to drop in d0ef03c.

- The `deferredReasoning` field and the post-tool thinking flush in `closeOpenToolCalls()` are gone.
- `reasoning_content` arriving while any tool_use is open is now silently dropped. The transcript stays `thinking@0 → tool_use@1`, no third block.
- Pre-tool thinking and the tool_use contents are unchanged.

Spec cases that previously pinned the deferred flush were flipped to assert the drop:
- "defers reasoning_content arriving during a tool_use..." → "drops reasoning_content arriving during an open tool_use to keep the tool_use block contiguous".
- "keeps a fragmented tool_call... reasoning_content interleaves..." now also asserts no `thinking_delta` carries the dropped fragment and only two `content_block_start` events fire.

Tests: `anthropic-messages-adapter` 40/40, full `routing/proxy` 1297/1297. End-to-end DeepSeek thinking via `/v1/messages` still emits `thinking@0 → text@1` in the expected order.

guillaumegay13:
Thanks @kyya, merged!

PR #1917, opened 2026-05-14
Fix provider metadata when duplicating agents

suainam:
## Summary

  Fix agent duplication failing when copied provider rows hit self-hosted PostgreSQL schemas that require non-null `user_providers.label`.

  This change preserves `label` and `priority` when duplicating `UserProvider` rows.

  ## Why

  `AgentDuplicationService` copied provider rows without `label`, which caused inserts to fail with a database error on schemas where `user_providers.label` is `NOT NULL`.

  ## Testing

  - Added regression coverage for duplicated provider metadata
  - Ran backend duplication unit tests
  - Ran full repo build
  - Verified in the running Docker Compose environment that duplicating an agent now succeeds

  ## Notes

  Backend e2e tests were not completed in this environment because the local test database was unavailable from the test runner.

---
## Summary by cubic
Fix agent duplication by preserving provider `label` and `priority` when copying `UserProvider` rows. This prevents insert failures on self-hosted PostgreSQL schemas that require a non-null `user_providers.label`, with regression tests added to ensure the metadata is kept.

<sup>Written for commit 59bb32e1a73f94155a9af030f91bf3f7902394dd. Summary will update on new commits.</sup>

### DIFF
--- a/.changeset/fix-duplicate-provider-label.md
+++ b/.changeset/fix-duplicate-provider-label.md
@@ -1,0 +1,5 @@
+---
+"manifest": patch
+---
+
+Preserve provider `label` and `priority` when duplicating agents so copied providers no longer fail database inserts on self-hosted PostgreSQL schemas that require non-null labels.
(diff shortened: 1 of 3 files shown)


SebConejo:
@suainam Superseded: the bug it describes can't happen on current main because duplication doesn't create new provider rows at all.

PR #2061, opened 2026-05-29
feat: subscription maximization pivot — provider pages, usage charts, agent provider access

SebConejo:
✨ What changed
- Three new provider pages (Subscriptions, BYOK, Local) replace the old provider management modal with full-page layouts including usage charts, sparklines, and connection detail pages
- Agent detail page restructured with 4 tabs: Routing, Providers, Guardrails, Settings
- Per-agent provider access control: toggle ON/OFF which providers each agent can use, with impact preview before disabling
- Global Overview page with skeleton loaders and provider category cards
- Usage analytics endpoint filtered by auth type and provider, with agent filter dropdown

💭 Why
The app was built around a single "connect provider" modal per agent. The new subscription maximization pivot treats providers as a user-level resource shared across agents, with per-agent access control. Users with multiple subscriptions need to see utilization, not just configuration.

👤 For users
- Provider management moved from per-agent modals to dedicated sidebar pages with usage charts and sparklines
- Each agent can selectively disable providers it shouldn't use
- Disabling a provider automatically cleans up affected routing assignments with a preview of what will change
- Model picker only shows models from enabled providers
- Guardrails tab replaces Limits with the same functionality

🔧 For operators
- New DB migration: `LiftProvidersToUserLevel` makes `user_providers.agent_id` nullable, creates `agent_provider_access` junction table
- New endpoints: `/api/v1/providers`, `/api/v1/provider-analytics`, `/api/v1/agents/:name/provider-access`
- `available-models` endpoint now filters by agent provider access

🧪 How to test
1. Navigate to Subscriptions/BYOK/Local pages: verify charts load, tables show connected providers with sparklines
2. Click "View details" on a connected provider: verify connection detail page with chart and agents table
3. Go to agent Providers tab: toggle OFF a provider, verify impact modal shows affected tiers
4. Go to Routing tab: verify disabled provider models don't appear in model picker
5. Verify "All providers disabled" empty state appears when all toggles are OFF

---
## Summary by cubic
Provider management is now user‑scoped with global provider/custom‑provider pages, stacked token/message/cost charts, rate‑limit snapshots, and a global Playground. UI polish extends to harness routes and icons, improves custom provider display on Overview, adds an “Add custom provider” button on BYOK, and fixes a focus bug in the provider key form.

- **New Features**
  - Global providers: `GET /api/v1/providers` lists user‑level connections with deduped usage/cost/model counts and last used; detail/analytics support `auth_type`/`provider` filters.
  - Analytics: `GET /api/v1/provider-analytics` plus per‑agent timeseries (`/overview/per-agent-timeseries`, `/overview/per-agent-message-timeseries`, `/overview/per-agent-cost-timeseries`) for tokens, messages, and cost (tab order: Cost → Messages → Token usage). Connection detail adds agents, models, recent messages.
  - Per‑agent access: `GET/PUT/DELETE /api/v1/agents/:name/provider-access` to toggle provider access per agent; routing/model pickers respect access with impact preview.
  - Custom providers: user‑scoped via `GET/POST /api/v1/custom-providers`; names/logos resolved with letter avatar fallback and a “Custom” badge; BYOK page now includes “Add custom provider”.
  - Rate limits: `GET /api/v1/rate-limits` returns header‑captured snapshots per provider/auth/key via `ProviderRateLimit`.
  - Global Playground at `/playground`; Messages page upgrades (Agent column, enriched provider filter); onboarding empty states and collapsible HARNESSES sidebar with add button. Routes renamed to `/harnesses` (was `/agents`) with updated back button label; platform icons added to the Messages Agent column; sidebar header padding tweaked.

- **Bug Fixes**
  - Migrations: lift `user_providers` to user scope with `agent_provider_access` junction table (keep `agent_id` nullable); lift `custom_providers` to user scope and drop `agent_id`; add `provider_rate_limits` table; label collisions are relabeled (not deleted) to avoid key loss; migrated connections are preserved and named by agent.
  - Routing/tiers: backfill/recalculation use `user_id`; provider removal recalculates routing across all of a user’s agents; cleanup no longer clears routes on other providers with similar model names; available‑models filters by access; cache splits agent‑ vs user‑scoped data with proper invalidation; fix routing‑cache invalidateAgent listener arg.
  - OAuth/keys: reconnect dedup and provider list dedup ensure a single logical connection per key; refresh is single‑flight keyed by `provider:user:label`; controllers use user‑scoped key lookups to avoid duplicate refreshes across agents.
  - Custom providers and access: case‑insensitive name conflicts return 409; lookups strictly user‑scoped; lists refetch after manage modal; cache clears on auth change. Agent provider access controller is user‑scoped with proper 404/409 errors; signatures/tests updated to thread `userId` across discovery, routing, and tier recalculation.
  - UI: load provider‑modal styles on provider pages; hide Manage button on disconnected connections; auto‑navigate to the correct provider type page when connecting; guard Playground history calls when no harnesses; Overview provider logos/text aligned and Type column normalized; ProviderKeyForm focus bug fixed; OpenClaw logo updated (`openclaw.svg`) and sidebar divider removed.
  - Dev tooling: add `.superset` setup/run/teardown scripts for one‑command local dev with Docker PostgreSQL and parallel backend/frontend.

<sup>Written for commit 63d57a52520b0baabd21c5deee8add989493be7b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2061?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


guillaumegay13:
Found a blocker in the `LiftProvidersToUserLevel` migration: it crashes on boot for anyone who had the same provider key connected to more than one agent (the normal case before this PR, since providers were per-agent).

Step 4 dedups by `api_key_encrypted`, but keys use a random IV so the same key has different ciphertext on each row, so the duplicates never merge. Step 6 then builds the unique index on `LOWER(label)` and hits the unmerged dupes.

Reproduced on a scratch DB with two "Default" anthropic rows across two agents:
```
could not create unique index "IDX_user_providers_user_provider_auth_label"
detail: Key (user_id, provider, auth_type, lower(label))=(..., default) is duplicated
```
Migrations run in a transaction on boot, so the app never starts and every restart retries the same failure.

Fix: dedup on `(user_id, provider, auth_type, LOWER(label))` instead of the ciphertext, keeping the lowest-priority row and reassigning its access entries. I have a tested patch plus a regression spec covering this exact two-ciphertext case, will push it up.

guillaumegay13:
Pushed the fix plus two related ones in #2104 (targets this branch): the migration dedup, a tier auto-assign backfill that broke on the user-level move (was filtering providers by the now-null agent_id), and an IDOR on the provider-access enable endpoint. Tests included.

PR #2104, opened 2026-06-03
fix: provider-lift migration dedup, tier backfill, and access IDOR

guillaumegay13:
Fixes three issues introduced by lifting providers to user level in #2061. Targets `coal-borogovia` so it folds into that PR.

### 1. Migration crashes on boot (blocker)
`LiftProvidersToUserLevel` dedups on `api_key_encrypted`, but keys use a random IV so the same key has different ciphertext on every row. Duplicates never merge, and the new unique index on `LOWER(label)` then fails for anyone who connected one provider to more than one agent (the normal case before this PR). Migrations run in a transaction on boot, so the app never starts.

Reproduced on a scratch DB: two "Default" anthropic rows across two agents, `CREATE UNIQUE INDEX` aborts with `Key (user_id, provider, auth_type, lower(label))=(..., default) is duplicated`.

Fix: dedup on `(user_id, provider, auth_type, LOWER(label))`, the same tuple the index enforces. Verified the migration now succeeds and merges the rows while keeping both agents' access.

### 2. Tier auto-assign backfill silently dead
`tier.service.getTiers` looked up active providers by `agent_id`, which is NULL after the lift, so `recalculate()` never ran and new tier slots stayed empty. Query by `user_id`.

### 3. IDOR on provider access
`enable()` granted an agent access to any `userProviderId` without checking it belonged to the caller, and there's no FK on the junction table to catch it. Scope the lookup to the user (matches the sibling `disable`/`impact` handlers).

### Tests
- New migration spec pinning the dedup tuple and the dedup-before-index order.
- New `agent-provider-access.controller` spec (the controller had none), 100% coverage, including the IDOR guard.
- New `tier.service` regression test asserting the `user_id` lookup.

---
## Summary by cubic
Fixes a boot-crashing migration, restores tier auto-assign backfill, and closes an access-control gap after lifting providers to user scope. Migrations now complete without data loss and provider access stays limited to the owner.

- **Bug Fixes**
  - Migration: relabel colliding rows on (user_id, provider, auth_type, LOWER(label)) before creating the unique index; never delete rows, so no keys are lost.
  - Tier backfill: query active providers by user_id so getTiers auto-assigns models into new slots.
  - Access control: enable() now verifies the provider belongs to the caller, preventing cross-user access (IDOR).
  - Tests: added specs for migration relabeling/order, controller IDOR guard, and tier.service user_id lookup.

<sup>Written for commit 1db63fc80821592746290088f1fb5dadf1d7106a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2104?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/backend/src/database/migrations/1791000000000-LiftProvidersToUserLevel.spec.ts
+++ b/packages/backend/src/database/migrations/1791000000000-LiftProvidersToUserLevel.spec.ts
@@ -42,29 +42,37 @@ describe('LiftProvidersToUserLevel1791000000000', () => {
     );
   });
 
-  it('dedups on the SAME tuple the unique index enforces, not on the encrypted key', async () => {
+  it('disambiguates colliding rows by relabeling on the index tuple, never on the encrypted key', async () => {
     await migration.up(queryRunner);
 
-    const dedup = queries.find((q) => q.includes('HAVING COUNT(*) > 1'));
-    expect(dedup).toBeDefined();
+    const relabel = queries.find((q) => q.includes('ROW_NUMBER() OVER'));
+    expect(relabel).toBeDefined();
     // Regression guard: keys are encrypted with a random IV, so the same key
-    // yields different ciphertext per row. Grouping on api_key_encrypted would
-    // never collapse those duplicates and the unique index in step 6 would then
-    // fail on boot. The dedup MUST group on (user_id, provider, auth_type,
+    // yields different ciphertext per row. Partitioning on api_key_encrypted
+    // would miss real collisions and the unique index in step 6 would fail on
+    // boot. The relabel MUST partition on (user_id, provider, auth_type,
     // LOWER(label)) — the exact tuple the new index keys on.
-    expect(dedup).toContain('GROUP BY "user_id", "provider", "auth_type", LOWER("label")');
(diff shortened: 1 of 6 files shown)


guillaumegay13:
Updated the migration fix after a second-opinion review flagged a flaw in the first pass. The original version deduped label collisions by deleting all but one row, which silently dropped a real key when two agents held different keys under the same label (random-IV encryption means SQL can't tell that apart from the same key on N agents). Switched to relabeling instead: every key is preserved, colliding rows get their id appended to the label, and the access rows stay correct. Verified both cases (same key on two agents, and different keys same label) migrate cleanly with no rows lost.

guillaumegay13:
Folded these two commits directly into the PR branch (coal-borogovia / #2061) since they belong there, so this stacked PR is redundant. The migration relabel fix, tier backfill fix, and access IDOR fix are now part of #2061.

PR #2110, opened 2026-06-03
fix: use shared OAuth credentials in Playground

guillaumegay13:
## Summary

- use the shared routing OAuth credential helpers from `oauth-credentials.ts` in Playground
- unwrap subscription OAuth credential blobs before Playground forwards provider calls
- re-read the latest persisted OAuth blob after preflight refresh so a later forced 401 retry does not use stale raw credentials
- add Playground parity with proxy routing by forcing one OAuth refresh/retry after an upstream 401
- wire OAuth services into the Playground module and cover the OAuth forwarding/retry paths

## Root Cause

Playground was reading the stored subscription credential directly and forwarding the persisted OAuth JSON blob as the bearer token. Normal proxy routing now has shared OAuth credential handling; Playground should use that same component instead of owning fallback-specific helper exports.

## Validation

- npm --workspace manifest-shared run build
- npm --workspace manifest-backend test -- playground.service.spec.ts --runInBand
- npm --workspace manifest-backend run build
- npx prettier --check packages/backend/src/playground/playground.service.ts packages/backend/src/playground/playground.module.ts packages/backend/src/playground/playground.service.spec.ts
- npx eslint packages/backend/src/playground/playground.service.ts packages/backend/src/playground/playground.module.ts packages/backend/src/playground/playground.service.spec.ts
- git diff --check
- npx changeset status --since=upstream/main

Related to mnfst/modelparams.dev#40.

### DIFF
--- a/packages/backend/src/playground/playground.module.ts
+++ b/packages/backend/src/playground/playground.module.ts
@@ -8,6 +8,7 @@ import { CommonModule } from '../common/common.module';
 import { ModelPricesModule } from '../model-prices/model-prices.module';
 import { RoutingCoreModule } from '../routing/routing-core/routing-core.module';
 import { ProxyModule } from '../routing/proxy/proxy.module';
+import { OAuthModule } from '../routing/oauth/oauth.module';
 import { PlaygroundController } from './playground.controller';
 import { PlaygroundService } from './playground.service';
 import { PlaygroundHistoryService } from './playground-history.service';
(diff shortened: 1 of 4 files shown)


guillaumegay13:
good catch — fixed in 3c9e467. MiniMax subscription requests now build the `minimax-subscription` endpoint override from the OAuth `resource_url` (same as the proxy via `normalizeMinimaxSubscriptionBaseUrl`), so they hit the right region instead of the default. invalid resource URLs are ignored with a warning. added tests for the region override + invalid-URL fallback.

PR #2113, opened 2026-06-04
fix: prevent OAuth subscription tokens from being invalidated on refresh (#2012)

### LINKED ISSUE #2012: Add "Renew Auth Token" / re-authenticate action for existing OAuth provider accounts
## Feature request

Please add a "Renew Auth Token" / "Re-authenticate" action for existing OAuth-based provider accounts, especially OpenAI/Codex subscription auth.

## Problem

OpenAI/Codex OAuth authentication appears to expire after a limited time, e.g. around 30 days.

Currently, when the auth expires, there does not seem to be a way to renew the login for an existing provider account. The workaround is to delete the provider account completely and then add it again.

That has a few downsides:

- existing routing settings can be lost or need to be recreated
- agent-specific provider/account selection needs to be configured again
- the process is error-prone when multiple agents use similar provider setups
- it feels like credential rotation, but behaves like deleting and recreating infrastructure

## Desired behavior

For an existing OAuth provider account, Manifest should offer a button such as:

- "Renew Auth Token"
- "Re-authenticate"
- "Refresh login"

This should start the same OAuth/login flow again, but update the existing provider account instead of creating a new one or requiring deletion first.

## Expected UX

Example flow:

1. User opens an existing OpenAI/Codex OAuth provider account
2. User clicks "Renew Auth Token"
3. Manifest opens the OAuth/device/login flow again
4. On success, Manifest updates the stored auth credentials for the existing provider account
5. Existing routing, account label, default status, and agent references stay intact

## Why this matters

OAuth/subscription auth is not a one-time setup if tokens expire periodically. Users should be able to rotate or refresh credentials without breaking their existing routing setup.

## Related

This would complement a central/shared provider credential vault, because shared OAuth credentials would still need a safe renewal flow.

guillaumegay13:
## Problem (fixes #2012)

OAuth subscription tokens are stored encrypted in `user_providers.api_key_encrypted` as a blob `{ t: access, r: refresh, e: expiry, u?: resource }`, and the proxy lazily refreshes the access token before forwarding. Providers like **OpenAI now issue rotating refresh tokens** — every successful refresh invalidates the old refresh token and returns a new one (verified live: `rt_…` 90-char → `rt.1.AAC…` 343-char; reusing the old one returns `401 refresh_token_invalidated`).

The previous `parse blob → refresh(blob.r) → upsert` path in each `unwrapToken()` had two fatal races under rotation:

1. **Concurrent requests.** An agent fires many parallel requests against an expired token. Each parsed its own (stale) blob and refreshed with the **same** old refresh token — the first rotated it, the rest got `refresh_token_invalidated`. A request arriving just after a refresh finished could still hold the pre-refresh blob (from the routing cache) and refresh again with the now-dead token.
2. **Persist-after-rotate failure.** If the DB write failed *after* the provider already rotated, the only copy of the new refresh token was lost while the stored one was already dead → the account is bricked until the user deletes and recreates it (exactly #2012). The old `catch` just logged and returned `null`, silently bricking it.

This was reproduced against production data: expired accounts returned `refresh_token_invalidated`, while accounts whose stored token hadn't been clobbered refreshed fine from a plain script.

## Fix

A single-flight **OAuth refresh coordinator** (`oauth/core/oauth-refresh-coordinator.ts`), wired into every subscription OAuth provider (OpenAI/Gemini via the shared base, plus Anthropic, MiniMax, xAI, Kiro):

- **Coalesce** concurrent refreshes for the same credential onto one in-flight promise, keyed by `provider:user:agent:label`. Different credentials never block each other.
- **Re-read the freshest blob from the DB** (bypassing the routing cache) before refreshing. If another request already refreshed it, return that and skip the redundant rotation. The fresh refresh token — not the caller's stale one — is the refresh source. New helper: `ProviderService.getFreshSubscriptionCredential()`.
- **Persist with retries**, and never refresh again on a persist failure (the old token is already dead — retrying the same rotated blob is the only safe move).

### Scope / follow-up

This makes refresh correct for a **single backend instance** (today's deployment). Two replicas refreshing the same credential at the same instant is intentionally out of scope — that needs an optimistic compare-and-set on persist or a DB lease, noted as a follow-up in the coordinator's header comment.

Note: this prevents healthy accounts from being bricked going forward; accounts whose refresh token OpenAI has **already** invalidated still require a reconnect.

## Tests

- New `oauth-refresh-coordinator.spec.ts` (100% coverage): coalescing, fresh-DB-copy reuse, unparseable-fresh-value fallback, persist retry/exhaustion, early-refresh skew, key formatting.
- New `ProviderService.getFreshSubscriptionCredential` cases (row found/decrypt, missing row, no credential, decrypt failure, label defaulting).
- Added MiniMax cases covering the new blob-parse helper.
- Full backend suite green (6058 tests).

---
## Summary by cubic
Prevents OAuth subscription tokens from being invalidated during refresh by coordinating refreshes and safely persisting rotated tokens. Fixes #2012 and stops accounts from getting bricked with rotating refresh tokens (e.g., OpenAI).

- **Bug Fixes**
  - Added a single-flight refresh coordinator that coalesces concurrent refreshes per credential (`provider:user:agent:label`) for a single backend instance.
  - Re-reads the freshest credential from the DB before refreshing and skips refresh if it’s already valid; uses the latest refresh token.
  - Persists rotated tokens with retries and never re-rotates on persist failures.
  - Integrated into all subscription OAuth providers (OpenAI/Gemini via the shared base; Anthropic, MiniMax, xAI, Kiro) and added `ProviderService.getFreshSubscriptionCredential()` for cache-bypassing reads.
  - Updated `ProviderService.getFreshSubscriptionCredential()` to match labels case-insensitively so fresh reads work even when label casing differs.

<sup>Written for commit 757c4ef764a06fb258027bad7224caa5a7089d1b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2113?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/oauth-refresh-coordinator.md
+++ b/.changeset/oauth-refresh-coordinator.md
@@ -1,0 +1,5 @@
+---
+"manifest": patch
+---
+
+Fix OAuth subscription tokens getting permanently invalidated (#2012). Providers like OpenAI now rotate refresh tokens on every refresh, so the previous "refresh then persist" path could brick an account when parallel proxy requests refreshed the same credential at once, or when the DB write failed after the provider had already rotated. Lazy refreshes are now coordinated per credential: concurrent refreshes coalesce into a single round-trip, the freshest token is re-read from the database before refreshing, and the rotated token is persisted with retries. Applies to all subscription OAuth providers (OpenAI, Gemini, Anthropic, MiniMax, xAI, Kiro).
--- a/packages/backend/src/routing/minimax-oauth.service.spec.ts
+++ b/packages/backend/src/routing/minimax-oauth.service.spec.ts
@@ -20,6 +20,7 @@ describe('MinimaxOauthService', () => {
       upsertProvider: jest.fn().mockResolvedValue({ provider: {}, isNew: true }),
       recalculateTiers: jest.fn().mockResolvedValue(undefined),
       nextOAuthLabel: jest.fn().mockResolvedValue(undefined),
+      getFreshSubscriptionCredential: jest.fn().mockResolvedValue(null),
     } as unknown as jest.Mocked<ProviderService>;
 
     configService = {
--- a/packages/backend/src/routing/oauth/anthropic/anthropic-oauth.service.security.spec.ts
+++ b/packages/backend/src/routing/oauth/anthropic/anthropic-oauth.service.security.spec.ts
@@ -22,11 +22,18 @@ function createProviderService() {
   const upsertProvider = jest.fn().mockResolvedValue({ provider: { id: 'p1' } });
   const recalculateTiers = jest.fn().mockResolvedValue(undefined);
   const nextOAuthLabel = jest.fn().mockResolvedValue(undefined);
+  const getFreshSubscriptionCredential = jest.fn().mockResolvedValue(null);
   return {
-    svc: { upsertProvider, recalculateTiers, nextOAuthLabel } as unknown as ProviderService,
+    svc: {
+      upsertProvider,
+      recalculateTiers,
+      nextOAuthLabel,
+      getFreshSubscriptionCredential,
+    } as unknown as ProviderService,
     upsertProvider,
     recalculateTiers,
     nextOAuthLabel,
+    getFreshSubscriptionCredential,
   };
 }
 
--- a/packages/backend/src/routing/oauth/anthropic/anthropic-oauth.service.spec.ts
+++ b/packages/backend/src/routing/oauth/anthropic/anthropic-oauth.service.spec.ts
@@ -27,11 +27,18 @@ function createProviderService() {
   const upsertProvider = jest.fn().mockResolvedValue({ provider: { id: 'p1' } });
   const recalculateTiers = jest.fn().mockResolvedValue(undefined);
   const nextOAuthLabel = jest.fn().mockResolvedValue(undefined);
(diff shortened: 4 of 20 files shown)


guillaumegay13:
Thanks — fixed in 757c4ef. `getFreshSubscriptionCredential` now matches the label case-insensitively (scanning the agent/provider subscription rows and comparing `label.toLowerCase()`), consistent with the `LOWER(label)` unique index and the rest of the label handling. Added a case-insensitivity test.

guillaumegay13:
shipped 🚀 context for anyone landing here:

OpenAI rotates refresh tokens on every refresh - the old path could double-refresh under parallel requests and trip OpenAI's reuse-detection, invalidating the whole token family. this coordinates refreshes per account (single-flight + fresh db read + persist-with-retries) so healthy accounts don't get bricked.

note: it's preventative. accounts OpenAI already invalidated need a one-time reconnect - it can't un-revoke them. multi-replica is out of scope (follow-up: optimistic CAS / db lease).

PR #2178, opened 2026-06-10
fix: make model routing user-controlled

guillaumegay13:
### ✨ What changed
- Provider connects and discovery no longer create or refresh routes.
- Provider/key disconnect blocks while any user route still uses it.
- Disconnect still removes provider access from agents when no routes depend on it.
- Routing resolution and UI read only explicit override routes.

### 💭 Why
Automatic routing changed model choices behind the user's back. Provider lifecycle changes should not mutate routes.

### 👤 For users
Connecting a provider grants access only. Users choose every model route themselves.

### 📝 Notes
Message tab fixes moved to #2181.


SebConejo:

PR #2179, opened 2026-06-10
fix: support large proxy image payloads

### LINKED ISSUE #2126: OpenAI-compatible vision requests with base64 images fail above 1MB body limit and return misleading 404
## Description

OpenAI-compatible vision requests using `image_url` with a base64 data URL fail when the JSON request body is slightly above ~1MB.

This affects clients such as OpenWebUI when sending normal image attachments to Manifest, because OpenWebUI converts uploaded images to `data:image/...;base64,...` inside the chat completion payload.

The failure response is also misleading: instead of returning `413 Payload Too Large` or a clear body-size error, Manifest returns:

```json
{
  "message": "Cannot POST /v1/chat/completions",
  "error": "Not Found",
  "statusCode": 404
}
```

## Environment

* Manifest version: `<insert version>`
* Endpoint used: `/v1/chat/completions`
* Client: direct `curl.exe` and OpenWebUI
* Provider selected by Manifest: `gpt-5.4` / OpenAI route
* OS used for reproduction: Windows PowerShell

## What works

A normal text-only request works:

```powershell
curl.exe -N -v --connect-timeout 30 --max-time 120 "https://<manifest-host>/v1/chat/completions" -H "Authorization: Bearer <REDACTED>" -H "Content-Type: application/json" -H "Expect:" --data-binary "@body.json"
```

Example response headers include:

```text
HTTP/1.1 200 OK
Content-Type: text/event-stream
X-Manifest-Model: gpt-5.4
X-Manifest-Provider: openai
```

A small base64 image request also works.

Example body size:

```text
Content-Length: 16814 bytes
```

Manifest responds correctly and the model describes the image.

## What fails

A larger base64 image request fails when the body is slightly above 1MB.

Example body size:

```text
Content-Length: 1049686 bytes
```

Response:

```text
HTTP/1.1 404 Not Found
```

Body:

```json
{
  "message": "Cannot POST /v1/chat/completions",
  "error": "Not Found",
  "statusCode": 404
}
```

This is very close to the 1 MiB threshold:

```text
1 MiB = 1048576 bytes
failing request = 1049686 bytes
```

## Reproduction

Create a minimal text request:

```powershell
@'
{"model":"auto","stream":true,"messages":[{"role":"user","content":"Rispondi solo: ok"}]}
'@ | Set-Content -Encoding UTF8 body.json
```

Send it:

```powershell
curl.exe -N -v --connect-timeout 30 --max-time 120 "https://<manifest-host>/v1/chat/completions" -H "Authorization: Bearer <REDACTED>" -H "Content-Type: application/json" -H "Expect:" --data-binary "@body.json"
```

Result: `200 OK`.

Then generate a JSON payload with a base64 image slightly above 1MB:

```python
import base64
import json
import struct
import sys
import zlib
from pathlib import Path


def chunk(tag: bytes, data: bytes) -> bytes:
    return (
        struct.pack(">I", len(data))
        + tag
        + data
        + struct.pack(">I", zlib.crc32(tag + data) & 0xFFFFFFFF)
    )


def make_red_png(width: int, height: int) -> bytes:
    raw = b"".join(
        b"\x00" + (b"\xff\x00\x00" * width)
        for _ in range(height)
    )

    return (
        b"\x89PNG\r\n\x1a\n"
        + chunk(
            b"IHDR",
            struct.pack(">IIBBBBB", width, height, 8, 2, 0, 0, 0),
        )
        + chunk(b"IDAT", zlib.compress(raw, level=0))
        + chunk(b"IEND", b"")
    )


size = int(sys.argv[1]) if len(sys.argv) > 1 else 512
png = make_red_png(size, size)

data_url = "data:image/png;base64," + base64.b64encode(png).decode("ascii")

payload = {
    "model": "auto",
    "stream": True,
    "stream_options": {
        "include_usage": True
    },
    "messages": [
        {
            "role": "user",
            "content": [
                {
                    "type": "text",
                    "text": "Cosa vedi in questa immagine? Rispondi in italiano."
                },
                {
                    "type": "image_url",
                    "image_url": {
                        "url": data_url
                    }
                }
            ]
        }
    ]
}

body = json.dumps(payload, separators=(",", ":"))
Path("payload.json").write_text(body, encoding="utf-8")

print("PNG bytes:", len(png))
print("Data URL length:", len(data_url))
print("JSON body bytes:", len(body.encode("utf-8")))
```

Run:

```powershell
python .\make_manifest_payload.py 512
```

Then send:

```powershell
curl.exe -N -v --connect-timeout 30 --max-time 180 "https://<manifest-host>/v1/chat/completions" -H "Authorization: Bearer <REDACTED>" -H "Content-Type: application/json" -H "Expect:" --data-binary "@payload.json"
```

Result:

```json
{
  "message": "Cannot POST /v1/chat/completions",
  "error": "Not Found",
  "statusCode": 404
}
```

## Expected behavior

Manifest should support OpenAI-compatible vision payloads with base64 images larger than 1MB, or at least allow the request body size limit to be configured.

If the request body is too large, the API should return a clear error such as:

```text
413 Payload Too Large
```

instead of:

```text
404 Cannot POST /v1/chat/completions
```

## Suspected cause

The backend appears to configure Express body parsing with a hardcoded 1MB limit:

```ts
expressApp.use(express.json({ limit: '1mb' }));
expressApp.use(express.urlencoded({ extended: true, limit: '1mb' }));
```

This is too small for OpenAI-compatible multimodal requests, because normal image uploads are commonly encoded as base64 data URLs inside JSON.

A 1.29MB JPEG becomes roughly 1.7MB once base64 encoded, before accounting for the rest of the JSON payload.

## Suggested fix

Make the body size limit configurable, for example:

```text
MANIFEST_BODY_LIMIT=25mb
```

or increase the default to a more vision-friendly value such as `25mb` or `50mb`.

Also, return a proper `413 Payload Too Large` error when the body parser rejects the request due to size.

### Failing request: body slightly above 1MB

```text
curl.exe -N -v --connect-timeout 30 --max-time 180 "https://<manifest-host>/v1/chat/completions" -H "Authorization: Bearer <REDACTED>" -H "Content-Type: application/json" -H "Expect:" --data-binary "@payload.json"

> POST /v1/chat/completions HTTP/1.1
> Host: <manifest-host>
> Content-Type: application/json
> Content-Length: 1049686
>
* upload completely sent off: 1049686 bytes

< HTTP/1.1 404 Not Found
< Content-Length: 83
< Content-Type: application/json; charset=utf-8

{"message":"Cannot POST /v1/chat/completions","error":"Not Found","statusCode":404}
```

This request is only slightly above 1 MiB:

```text
1 MiB = 1048576 bytes
failing request = 1049686 bytes
```

### Successful request: small base64 image

Generated payload:

```text
Generating 64x64 red PNG...
PNG bytes: 12420
Data URL length: 16582
JSON body bytes: 16814
```

Request:

```text
curl.exe -N -v --connect-timeout 30 --max-time 180 "https://<manifest-host>/v1/chat/completions" -H "Authorization: Bearer <REDACTED>" -H "Content-Type: application/json" -H "Expect:" --data-binary "@payload.json"

> POST /v1/chat/completions HTTP/1.1
> Host: <manifest-host>
> Content-Type: application/json
> Content-Length: 16814
>
* upload completely sent off: 16814 bytes

< HTTP/1.1 200 OK
< Content-Type: text/event-stream
< X-Manifest-Model: gpt-5.4
< X-Manifest-Provider: openai
< X-Manifest-Response-Mode: stream
```

Relevant response excerpt:

```text
data: {"model":"gpt-5.4","choices":[{"delta":{"content":"Vedo un’immagine completamente rossa, uniforme, senza altri elementi visibili."}}]}
data: [DONE]
```

guillaumegay13:
### ✨ What changed
- `/v1/*` accepts 512 MB OpenAI-compatible proxy bodies.
- Regular Manifest API routes stay at 1 MB.
- Inline image data is redacted for routing and recordings.
- Proxy body admission returns 413, 429, or 411 before parsing.

### 💭 Why
OpenAI-compatible clients can send base64 image data URLs. Manifest was rejecting them at 1 MB and could accept too many large bodies at once.

### 👤 For users
Large inline vision requests no longer fail at Manifest's parser.

### 🔧 For operators
No env vars or migrations.

### 📝 Notes
Closes #2126.
Local smoke: 1.4 MB inline image reached a fake OpenAI provider while recordings stored only the placeholder; `/api` stayed at 1 MB; chunked and >512 MB proxy bodies were rejected before parsing.
Local checks: focused backend Jest, backend build/lint, shared build, changeset status, `git diff --check`.

### DIFF
--- a/.changeset/proxy-inline-image-body-limit.md
+++ b/.changeset/proxy-inline-image-body-limit.md
@@ -1,0 +1,5 @@
+---
+'manifest': patch
+---
+
+Support large OpenAI-compatible inline image requests on `/v1/*` with route-scoped body parsing, clear body-size errors, and redacted inline image data for routing and message recordings.
--- a/packages/backend/src/common/middleware/body-parser-limits.spec.ts
+++ b/packages/backend/src/common/middleware/body-parser-limits.spec.ts
@@ -1,0 +1,121 @@
+import express, { Request, Response } from 'express';
+import { EventEmitter } from 'events';
+import request from 'supertest';
+import {
+  API_BODY_LIMIT,
+  PROXY_BODY_LIMIT,
+  PROXY_BODY_LIMIT_BYTES,
+  bodyParserErrorHandler,
+  createProxyBodyBudgetMiddleware,
+} from './body-parser-limits';
+
+function createParserTestApp() {
+  const app = express();
+  app.use('/v1', createProxyBodyBudgetMiddleware());
+  app.use('/v1', express.json({ limit: PROXY_BODY_LIMIT }));
+  app.use(express.json({ limit: API_BODY_LIMIT }));
+  app.use(bodyParserErrorHandler);
+  app.post('/v1/chat/completions', (req: Request, res: Response) => {
+    res.json({ contentLength: String(req.body.messages[0].content).length });
+  });
+  app.post('/api/v1/test', (req: Request, res: Response) => {
+    res.json({ contentLength: String(req.body.messages[0].content).length });
+  });
+  return app;
+}
+
+function createMockResponse() {
+  const res = new EventEmitter() as express.Response & EventEmitter;
+  res.status = jest.fn(() => res) as unknown as express.Response['status'];
+  res.json = jest.fn(() => res) as unknown as express.Response['json'];
+  return res;
+}
+
+describe('body parser limits', () => {
+  it('allows OpenAI-compatible proxy bodies above the regular Manifest API limit', async () => {
+    const content = 'x'.repeat(1024 * 1024 + 1);
+    const payload = { messages: [{ role: 'user', content }] };
+
+    await request(createParserTestApp()).post('/v1/chat/completions').send(payload).expect(200);
+    const apiRes = await request(createParserTestApp())
+      .post('/api/v1/test')
+      .send(payload)
+      .expect(413);
+
+    expect(apiRes.body).toEqual({
+      message: 'Request body is too large',
+      error: 'Payload Too Large',
+      statusCode: 413,
+    });
+  });
+
+  it('returns JSON 400s for malformed proxy JSON', async () => {
+    const res = await request(createParserTestApp())
+      .post('/v1/chat/completions')
+      .set('Content-Type', 'application/json')
+      .send('{"messages":')
+      .expect(400);
+
+    expect(res.body).toEqual({
+      message: 'Invalid JSON request body',
+      error: 'Bad Request',
+      statusCode: 400,
+    });
+  });
+
+  it('limits the total Content-Length of proxy request bodies in flight', () => {
+    const middleware = createProxyBodyBudgetMiddleware();
+    const firstReq = {
+      method: 'POST',
+      headers: { 'content-length': String(PROXY_BODY_LIMIT_BYTES) },
+    } as express.Request;
+    const secondReq = {
+      method: 'POST',
+      headers: { 'content-length': '1' },
+    } as express.Request;
+    const firstRes = createMockResponse();
+    const secondRes = createMockResponse();
+    const firstNext = jest.fn();
+    const secondNext = jest.fn();
+
+    middleware(firstReq, firstRes, firstNext);
+    middleware(secondReq, secondRes, secondNext);
+
+    expect(firstNext).toHaveBeenCalledTimes(1);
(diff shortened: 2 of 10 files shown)


SebConejo:

PR #2181, opened 2026-06-10
fix: restore Message tab harness filtering and icons

guillaumegay13:
### ✨ What changed
- Harness View more carries `?agent=` into global Messages.
- Global Messages seeds its harness filter from the query param.
- The Harness column stays visible in the global table.
- Harness cells show the saved platform icon when metadata exists.

### 💭 Why
These Message tab affordances were lost while the stack was split.

### 👤 For users
Harness-level View more opens the global log already filtered to that harness.


SebConejo:

PR #2187, opened 2026-06-10
fix: route header docs link by current page

guillaumegay13:
### ✨ What changed
- Routes header Docs links by the current page.
- Maps Limits, Routing, and provider pages to their matching docs.
- Keeps the introduction docs as the fallback.

### 💭 Why
Production already sends users to relevant docs. The stacked UI should keep that behavior.

### 👤 For users
Docs opens the page-specific documentation instead of always opening the main docs page.

### 📝 Notes
Validation passed `Header.test.tsx` and `git diff --check`.


SebConejo:

PR #2188, opened 2026-06-10
fix: count legacy agent-name rows for limits

guillaumegay13:
### ✨ What changed
- Counts legacy usage rows with `agent_name` when `agent_id` is missing.
- Keeps the new `agent_id` match as the preferred consumption path.
- Keeps the fallback scoped to the same tenant.

### 💭 Why
Older production message rows can miss `agent_id`. Limits read those rows as zero usage.

### 👤 For users
Hard limits apply to existing message history again after the stack upgrade.

### 📝 Notes
Validation passed backend notification/limit tests, frontend Limits tests, builds, and `git diff --check`.


SebConejo:

PR #2191, opened 2026-06-10
fix: compute limit period boundaries in local time, not UTC

guillaumegay13:
Stacked on top of #2188.

## The bug

Token/cost **limits silently never block** (and threshold alerts never fire) whenever the server process runs in a **non-UTC timezone**. The hard-limit feature shipped in `dcfa4b6e` is affected on `main` too — a Railway/UTC container works by coincidence, but any self-hosted install in another zone gets non-functional limits.

## Root cause

`computePeriodBoundaries` / `computePeriodResetDate` build their window with `Date.UTC(...)` + `toSqlTimestamp` (which uses `toISOString()` → **UTC** wall-clock). But `agent_messages.timestamp` is `timestamp without time zone` written in the process's **local** time — the pg driver serialises JS Dates locally, and the rest of the analytics layer already compares against local cutoffs via `computeCutoff`/`formatLocalIso` (see the note in `postgres-sql.ts`).

On a non-UTC host the window's upper bound (`periodEnd` = UTC "now") lands **behind** the local-time rows by the TZ offset, so `getConsumption`'s `SUM(...)` matches ~0 rows. With `actual (≈0) < threshold`, `ProxyService.enforceLimits` never trips and `NotificationCronService` never alerts.

PR #2188 fixed a *different* limits bug (legacy null-`agent_id` rows), so this gap remained.

## Reproduction (UTC+2 host)

- Agent with a `tokens` limit of **1/day**, `action: both`, active.
- Agent had **522 tokens** recorded today.
- Buggy UTC boundaries: `WHERE timestamp >= '…00:00:00' AND < '…11:16:00'` (UTC) excludes every `13:xx` local row → consumption `0` → request **allowed (HTTP 200, real completion)**.
- After the fix: consumption `522` → request **blocked**:
  ```
  [🦚 Manifest M200] You hit your tokens limit: 522 used, 1/day allowed.
  ```

## The fix

- Add `toLocalSqlTimestamp` (local-time sibling of `toSqlTimestamp`) in `postgres-sql.ts`.
- Recompute both period helpers from **local** `Date` parts + `toLocalSqlTimestamp`.
- Update `period.util` / `postgres-sql` specs to assert local behaviour, plus a round-trip regression that has teeth on any non-UTC host (CI runs UTC, where the old code coincidentally passed and hid this).
- Make the `notification-cron` edge-case boundary assertions timezone-independent (they previously hard-coded UTC strings).

## Tests

- `period.util.spec.ts`, `postgres-sql.spec.ts`: 100% line/branch coverage on both changed files.
- `notification-cron.service.edge-cases.spec.ts`: updated, passing.
- Note: `baseline-cost.spec.ts` (`collectRoutedModelIds`) fails on this branch **before** this change too — pre-existing on the stack, unrelated.

---
## Summary by cubic
Fixes a timezone bug that made token/cost limits and threshold alerts fail on non-UTC servers by computing period windows in local time to match stored message timestamps. Hard limits now block correctly and alerts fire as expected.

- **Bug Fixes**
  - Added `toLocalSqlTimestamp` and used it for limit window boundaries and reset dates.
  - Rewrote `computePeriodBoundaries` and `computePeriodResetDate` to use local time (not UTC).
  - Updated tests and notification-cron assertions to be timezone-agnostic.

<sup>Written for commit c47c089c173e37d8edbc7d63aaaf6ec6e3939dc5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2191?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/limits-local-time-boundaries.md
+++ b/.changeset/limits-local-time-boundaries.md
@@ -1,0 +1,5 @@
+---
+"manifest": patch
+---
+
+Fix token/cost limits never blocking (or alerting) when the server's timezone isn't UTC. `computePeriodBoundaries`/`computePeriodResetDate` built their window in UTC, but `agent_messages.timestamp` rows are stored in the process's local time — so on a non-UTC host the window's upper bound sat behind the stored rows by the TZ offset and the consumption SUM read ~0, meaning hard limits silently never tripped and threshold alerts never fired. Boundaries are now computed in local time (matching `computeCutoff`), via a new `toLocalSqlTimestamp` helper.
--- a/packages/backend/src/common/utils/period.util.spec.ts
+++ b/packages/backend/src/common/utils/period.util.spec.ts
@@ -1,5 +1,9 @@
 import { computePeriodBoundaries, computePeriodResetDate } from './period.util';
 
+// Parse a local-time `YYYY-MM-DD HH:MM:SS` string (no tz suffix) back into a
+// Date using the same local-timezone interpretation the formatter used.
+const parseLocal = (s: string) => new Date(s.replace(' ', 'T'));
+
 describe('computePeriodBoundaries', () => {
   it('returns periodStart and periodEnd as formatted strings', () => {
     const { periodStart, periodEnd } = computePeriodBoundaries('day');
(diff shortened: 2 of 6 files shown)


SebConejo:

PR #2192, opened 2026-06-10
fix: show custom provider names in overview series and models

guillaumegay13:
### ✨ What changed
- Provider-grouped chart series remap `custom:<uuid>` keys to the provider's display name.
- The provider filter, legend, and tooltip pick the name up automatically.
- Recent Messages and Model usage run model names through `getModelDisplayName`, so `custom:<uuid>/model` slugs render as the bare model name.

### 💭 Why
A custom provider named "Hugging Face" showed up as `custom:4b580cbf-...` in the Overview filter, chart tooltip, and model tables.

### 👤 For users
Custom providers read by their name everywhere on the Overview, same as built-in providers.

---
## Summary by cubic
Shows custom provider names across the Overview and global Messages log. Replaces `custom:<uuid>` with provider names and strips `custom:<uuid>/model` slugs to show only the model.

- **Bug Fixes**
  - Remap provider-grouped chart series keys to provider display names; updates charts, filters, legends, and tooltips once custom provider data resolves.
  - In the global Messages log, fall back to the first available agent when no route agent is present so custom providers load and their name + icon render correctly; prevents `custom:<uuid>/…` labels.

<sup>Written for commit 39b247904a8dd558832877ab1dbfc2cc9eaadfbc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2192?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #2190, opened 2026-06-10
fix: strip native button chrome from chart stat tabs

guillaumegay13:
### ✨ What changed
- `.chart-card__stat--clickable` resets the UA button background, border, and font.
- Keeps the themed divider border between stats.

### 💭 Why
The Cost/Messages/Token usage stats render as `<button>` for a11y, but the user-agent default styles painted them as light gray cards on the dark theme. Seb's design (#2061) uses plain divs with no chrome.

### 👤 For users
Overview chart stat tabs match the card background in both themes again.

---
## Summary by cubic
Strip default browser button styles from chart stat tabs so they match the card background in both themes; keeps the themed divider border.

<sup>Written for commit 17c8a15d8c0adf6395ed24b95c98a1f01b2b1a8f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2190?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #2189, opened 2026-06-10
fix: target the visible provider key on edit and rename

guillaumegay13:
### ✨ What changed
- Add-key form no longer steals focus to the key field on every name keystroke.
- Updating a key value targets the visible key's label instead of the 'Default' row.
- Key rename rewrites pinned routes across all the user's agents, not just one.
- Mutated agents get their routing cache flushed after a rename.

### 💭 Why
Three multi-key bugs found while testing the stack: the name field was untypable past one character, key edits silently landed on the first key, and renames left stale pins that made the proxy fall back to the default key.

### 👤 For users
Naming a second API key works, and editing or renaming a key affects the key you're looking at.

### 📝 Notes
Focus fix ported from #2061. Backend (54) and frontend (48) suites pass.

---
## Summary by cubic
Fixes multi-key editing and renaming so updates apply to the visible key and renames relabel routes across all of a user’s agents. The add-key modal no longer steals focus, making the name field typeable.

- **Bug Fixes**
  - Frontend: Updates now include the visible key’s label to avoid writing to the `Default` row.
  - Frontend: The add-key modal focuses the API key field only on open (tracks `isOpen` via `on()`), preventing focus theft while typing the name.
  - Backend: Renaming a key relabels pinned routes across all of the user’s agents (tiers, specificity, and header tiers) and flushes each mutated agent’s routing cache.

<sup>Written for commit 5dcb1a470666adf77d03770be1f265e8f8c75893. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2189?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #2193, opened 2026-06-10
Stacked review: limits TZ fix + provider/chart/custom-name fixes (all-in-one)

guillaumegay13:
**Aggregate branch for reviewing/testing the whole batch together — not for direct merge.** Merge the individual PRs below (each is small and independently reviewable); this branch just bundles their latest tips so you can run everything in one instance.

## Stack

All four PRs branch off **#2188** (`fix/limits-regression`), which is itself part of the global-providers stack. Two independent forks (disjoint files, clean merge):

```
#2188  fix/limits-regression            (base of this batch)
 ├─ #2191  fix/limits-tz-boundaries       → fix/limits-regression
 └─ #2189  fix/provider-key-edit-bugs     → fix/limits-regression
     └─ #2190  fix/chart-stat-button-styles   → fix/provider-key-edit-bugs
         └─ #2192  fix/custom-provider-display-names → fix/chart-stat-button-styles
```

This branch = merge(`#2191`, `#2192`) → contains #2188 + #2189 + #2190 + #2191 + #2192.

### #2188 — count legacy agent-name rows for limits *(base)*
Limit consumption query also counts rows where `agent_id IS NULL` and matches on `agent_name`, so legacy/global-provider rows are included in token/cost totals.

### #2191 — compute limit period boundaries in local time, not UTC
**Token/cost limits never blocked (or alerted) on non-UTC servers.** `computePeriodBoundaries`/`computePeriodResetDate` built the window in UTC, but `agent_messages.timestamp` is stored in the process's local time — so `periodEnd` sat behind the rows by the TZ offset and the consumption `SUM` read ~0. Now computed in local time (matching `computeCutoff`) via a new `toLocalSqlTimestamp`. Repro/fix verified on a UTC+2 host: a 1-token/day cap now blocks a 522-token agent (`M200`). *(Bug also present on `main`; UTC containers worked by luck.)*

### #2189 — target the visible provider key on edit and rename
Add-key form no longer steals focus on each keystroke; key edits land on the visible key (not the `Default` row); rename rewrites pinned routes across **all** the user's agents (keys are user-global) and flushes each mutated agent's routing cache.

### #2190 — strip native button chrome from chart stat tabs
Overview chart stat tabs (Cost/Messages/Tokens) render as `<button>` for a11y; reset the UA background/border/font so they match the card in both themes.

### #2192 — show custom provider names (+ icon) instead of `custom:<uuid>`
- **Overview**: provider-grouped series/filter/legend/tooltip and model tables remap `custom:<uuid>` → provider display name.
- **Messages log** *(follow-up commit)*: the global "All harnesses" view had no route agent, so the custom-providers resource never ran and rows showed `custom:Custom/<model>` with a letter badge. Now falls back to the first agent (custom providers are user-global), resolving the provider name + logo (e.g. "Hugging Face") like every other provider.

## Testing
- Backend: `period.util`/`postgres-sql` specs 100% line/branch; `notification-cron` edge cases TZ-independent.
- Frontend: `MessageLog.test.tsx` 93 tests incl. a global-mode custom-provider test; `MessageLog.tsx` 100% line coverage.
- Verified live on a self-hosted instance: limits block correctly; custom provider name + icon render in Messages.
- Pre-existing unrelated red on the stack: `baseline-cost.spec.ts` (`collectRoutedModelIds`).

### DIFF
--- a/.changeset/limits-local-time-boundaries.md
+++ b/.changeset/limits-local-time-boundaries.md
@@ -1,0 +1,5 @@
+---
+"manifest": patch
+---
+
+Fix token/cost limits never blocking (or alerting) when the server's timezone isn't UTC. `computePeriodBoundaries`/`computePeriodResetDate` built their window in UTC, but `agent_messages.timestamp` rows are stored in the process's local time — so on a non-UTC host the window's upper bound sat behind the stored rows by the TZ offset and the consumption SUM read ~0, meaning hard limits silently never tripped and threshold alerts never fired. Boundaries are now computed in local time (matching `computeCutoff`), via a new `toLocalSqlTimestamp` helper.
--- a/.changeset/setup-modal-stale-platform.md
+++ b/.changeset/setup-modal-stale-platform.md
@@ -1,0 +1,5 @@
+---
+'manifest': patch
+---
+
+Fix the post-create "Set up harness" modal showing the full harness picker instead of the harness you just chose. AgentGuard now refetches the agent list when the viewed agent changes, so a newly created agent's platform reaches the setup modal (previously the stale, source-less list left the platform unset whenever the agent was created from the always-present sidebar while another agent was open).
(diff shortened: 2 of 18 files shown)


SebConejo:

PR #2213, opened 2026-06-11
fix: strip context_management and signed thinking blocks from Anthropic proxy requests

JosiahSiegel:
fix: strip `context_management` and signed thinking blocks from Anthropic proxy requests

**Problem 1 — `context_management` strip**

Clients sending requests via `POST /v1/messages` (Anthropic Messages API format) may include a `context_management` field — a legitimate Anthropic parameter for compaction that requires the `compact-2026-01-12` beta header. Manifest forwards this field to Anthropic unchanged, but sends `anthropic-version: 2023-06-01` without the compaction beta header, causing Anthropic to reject the request:

```json
{"type":"error","error":{"type":"invalid_request_error","message":"context_management: Extra inputs are not permitted"}}
```

**Fix 1**

Strips `context_management` from the outbound request body in `applyAnthropicMessagesMutations()`, the passthrough function used when both inbound and upstream are Anthropic Messages format. Follows the existing `OPENAI_ONLY_FIELDS` pattern: a `Set` of known-problematic field names that gets deleted from the spread result.

**Problem 2 — signed thinking blocks on model mismatch**

When Manifest routes an Anthropic Messages request to a model different from the one the client originally requested, signed `thinking` and `redacted_thinking` blocks from prior assistant turns carry signatures that are only valid for the original model. The upstream rejects with:

```json
{"type":"error","error":{"type":"invalid_request_error","message":"Invalid signature in thinking block"}}
```

**Fix 2**

Removes `thinking` and `redacted_thinking` content blocks from assistant messages and drops the top-level `thinking` parameter when the `stripThinkingBlocks` option is set — which happens automatically when the inbound model differs from the resolved upstream model.

**Testing**

- Added test: verifies `context_management` is stripped while other Anthropic-native fields (`top_k`, `stop_sequences`, `thinking`, `metadata`, `tool_choice`) pass through untouched.
- Added tests for `stripThinkingBlocks`: removes blocks, respects the flag, does not mutate the original body.
- All 134 existing anthropic-adapter tests pass.

---
## Summary by cubic
Prevents Anthropic proxy failures by stripping `context_management` when the compaction beta header isn’t sent and removing signed `thinking` blocks when routing to a different model. This stops upstream `invalid_request_error` and signature validation errors.

- **Bug Fixes**
  - For Anthropic Messages passthrough, delete `context_management` before forwarding.
  - When inbound and routed models differ, remove `thinking`/`redacted_thinking` blocks from assistant messages and drop the top-level `thinking` param.

<sup>Written for commit f79dab486b4cbd5b89bfbc3ac9fc83d468940b2c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2213?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/modern-wombats-accept.md
+++ b/.changeset/modern-wombats-accept.md
@@ -2,4 +2,12 @@
 "manifest": patch
 ---
 
-fix: strip `context_management` from Anthropic proxy requests to prevent `invalid_request_error`
+fix: strip `context_management` and signed thinking blocks from Anthropic proxy requests
+
+Strips `context_management` from Anthropic Messages proxy requests to prevent
+`invalid_request_error` when the compaction beta header isn't sent.
+
+When routing to a different Anthropic model than the client requested, removes
+signed `thinking` and `redacted_thinking` blocks from prior assistant turns
+whose signatures are only valid for the original model, preventing
+`Invalid signature in thinking block` upstream rejections.
--- a/.changeset/modern-wombats-accept.md
+++ b/.changeset/modern-wombats-accept.md
@@ -1,0 +1,5 @@
+---
+"manifest": patch
+---
+
+fix: strip `context_management` from Anthropic proxy requests to prevent `invalid_request_error`
--- a/packages/backend/src/routing/proxy/__tests__/anthropic-adapter.spec.ts
+++ b/packages/backend/src/routing/proxy/__tests__/anthropic-adapter.spec.ts
@@ -2319,6 +2319,88 @@ describe('Anthropic Adapter', () => {
         messages: [{ role: 'user', content: 'hi' }],
       });
     });
+
+    describe('stripThinkingBlocks', () => {
+      it('removes thinking and redacted_thinking blocks from assistant messages', () => {
+        const result = applyAnthropicMessagesMutations(
+          {
+            model: 'claude-sonnet-4-20250514',
+            max_tokens: 1024,
+            thinking: { type: 'enabled', budget_tokens: 4096 },
+            messages: [
+              { role: 'user', content: 'hello' },
+              {
+                role: 'assistant',
+                content: [
+                  { type: 'thinking', thinking: 'reasoning', signature: 'sig_abc' },
+                  { type: 'redacted_thinking', data: 'opaque' },
+                  { type: 'tool_use', id: 'toolu_1', name: 'foo', input: {} },
+                ],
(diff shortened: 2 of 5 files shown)


JosiahSiegel:
Closing — both fixes in this PR have been adopted upstream:

**1. `context_management` strip (lines 96-102 of `anthropic-adapter.ts` here)**
Landed upstream as part of [mnfst/manifest#2232](https://github.com/mnfst/manifest/pull/2232) (commit `abbf5741b` — *fix: inject cache_control for Anthropic subscription OAuth requests*). The `ANTHROPIC_UNSUPPORTED_FIELDS = new Set(['context_management'])` strip is now present in upstream's `applyAnthropicMessagesMutations`.

**2. `stripThinkingBlocks` on Anthropic model mismatch (lines 113-128 here)**
Adopted upstream and restructured: upstream removed the explicit `stripThinkingBlocks` option entirely and replaced it with a `shouldForwardAnthropicThinking(body.thinking, _model)` gate in `toAnthropicRequest` plus the thinking-block replay logic in `applyAnthropicMessagesMutations` (only prepends cached blocks when no thinking prelude already exists). The same "remove signed thinking blocks from prior assistant turns on model mismatch" behavior is preserved.

**Verification**: diff between `origin/main` and `upstream/main` on `packages/backend/src/routing/proxy/anthropic-adapter.ts` shows the local fork has only the two fixes from this PR, and upstream has the equivalent behavior (with the restructuring noted above). The other files in this PR (`provider-client.ts` plumbing, the spec additions) are also subsumed by upstream's restructure.

**Tracking**: the fork's housekeeping-overlay sync pipeline has been updated to drop the obsolete patches (`c1cbc3a84`). `fix/anthropic-proxy-combined` branch will be deleted from both `JosiahSiegel/manifest` and the local clone after this PR is closed.

Thanks @cubic for the clean review — identified by cubic.

JosiahSiegel:
Closing as superseded — both fixes adopted upstream (see comment above). Branch will be deleted.

PR #2220, opened 2026-06-12
fix: restore prompt-cache hits for openai-subscription

### LINKED ISSUE #2217: openai-subscription provider never gets prompt caching — missing session-id/thread-id headers and x-codex-turn-state replay
## Summary

Requests routed through the `openai-subscription` provider (`chatgpt.com/backend-api/codex/responses`) **never hit the prompt cache**: `usage.input_tokens_details.cached_tokens` is 0 on every request, even when the client sends identical 20K+-token prefixes back-to-back with a stable `prompt_cache_key`. For agentic tool loops this means every step re-pays the full prefix in latency and subscription quota.

## Measurement

Prod-shaped streaming request on `/v1/responses` (≈22.3K input tokens: large system prompt + 35 tool schemas), sent twice back-to-back with an identical prefix:

| Path | Cached tokens on retry |
|---|---|
| Manifest → openai-subscription (`gpt-5.3-codex-spark`) | **0** (also 0 with explicit `prompt_cache_key`) |
| Same prefix, direct OpenAI API key (`api.openai.com/v1/responses`) | **22,016 / 22,355 (98.5%)** |

I verified the 0 is genuine and not a reporting artifact: the usage passthrough is faithful (`chatgpt-adapter.ts:271` forwards `input_tokens_details.cached_tokens` as-is).

## Root cause (code refs at `3b61e1d`)

The body conversion is fine — `toNativeResponsesRequest` (`packages/backend/src/routing/proxy/responses-adapter.ts:157`) spreads the full body, and `CODEX_SUBSCRIPTION_UNSUPPORTED_PARAMS` (`responses-adapter.ts:147`) doesn't strip `prompt_cache_key`, so it reaches the backend. The problem is **routing affinity via headers**.

`chatgptSubscriptionHeaders` (`packages/backend/src/routing/proxy/provider-endpoints.ts:115`) sends only:

```ts
const chatgptSubscriptionHeaders = (apiKey: string) => ({
  Authorization: `Bearer ${apiKey}`,
  'Content-Type': 'application/json',
  originator: CODEX_CLI_ORIGINATOR,
  'user-agent': CODEX_CLI_USER_AGENT,
});
```

The real Codex CLI (openai/codex, `codex-rs`) does three more things on every request to this backend:

1. **Sends `session-id` and `thread-id` headers** — `codex-api/src/requests/headers.rs` (`build_session_headers`).
2. **Defaults `prompt_cache_key` to the stable thread id** — `core/src/client.rs` (`fn prompt_cache_key`, ~L363).
3. **Captures and replays the `x-codex-turn-state` sticky-routing token** for all requests within a turn — `core/src/client.rs` ~L231: *"The `x-codex-turn-state` sticky-routing token, which must be replayed for all requests within [the turn]"*.

That turn-state replay is what pins consecutive requests to the same backend shard so the prefix cache can hit. There are zero references to `session-id`, `thread-id`, or `x-codex-turn-state` anywhere in `packages/backend/src` — so every request lands on an arbitrary shard and the cache never warms.

Secondary: the Chat Completions → Responses conversion (`toResponsesRequest`, `chatgpt-adapter.ts`) rebuilds the body from a fixed field list and drops `prompt_cache_key` entirely, so `/v1/chat/completions` callers can't even forward a cache key.

## Suggested fix

For `openai-subscription` requests:

1. Capture `x-codex-turn-state` from upstream responses and replay it on subsequent requests within the same logical session/turn.
2. Send stable `session-id` / `thread-id` headers — derived from the caller's `prompt_cache_key` (or Manifest's own session identity) would be enough.
3. (Minor) forward `prompt_cache_key` on the Chat Completions conversion path too.

Happy to provide the probe script or re-run measurements against a fix.

guillaumegay13:
### ✨ What changed
- New `CodexSessionAffinity`: sends `session-id`/`thread-id` headers, injects a stable `prompt_cache_key` when missing, replays `x-codex-turn-state`.
- `ProviderClient.forward()` applies affinity only on `openai-subscription` (chat + native responses paths).
- `toResponsesRequest` gains opt-in `forwardPromptCacheKey`, enabled for `openai-subscription` and `openai-responses` only.

### 💭 Why
Codex's backend (`chatgpt.com/backend-api/codex/responses`) only serves cache hits when consecutive requests hit the same shard. The real CLI pins that affinity via session/thread headers, a default `prompt_cache_key`, and a replayed turn-state token. Manifest sent none of these, so every subscription request landed on a random shard and re-paid the full prefix. See #2217.

### 👤 For users
ChatGPT subscription routing now gets prompt-cache hits in tool loops, cutting repeated-prefix latency and subscription quota.

### 📝 Notes
- Fixes #2217.
- Turn-state is per-instance; in multi-replica deploys each replica converges on its own affinity, degrading to today's behavior at worst.
- Credential is only an in-memory map key, never hashed or sent. 5-min sliding TTL, 10k-entry cap, turn-state dropped on non-2xx.
- New `codex-session-affinity.spec.ts` (16 tests, 100% coverage) plus header and request-conversion tests. Full backend suite green.

---
## Summary by cubic
Restore prompt-cache hits for ChatGPT subscription by sending Codex session-affinity headers and preserving `prompt_cache_key` where supported. This cuts repeated prefix latency and subscription usage in tool loops. Fixes #2217.

- Bug Fixes
  - Added `CodexSessionAffinity`: sets `session-id`/`thread-id`, injects a stable per-session `prompt_cache_key` when missing, and replays `x-codex-turn-state`. Uses random memoized IDs (credential never hashed or sent), 5‑min sliding TTL with 10k cap; caps caller `prompt_cache_key` at 512 chars (longer treated as absent) to avoid memory amplification; drops turn-state on non‑2xx, keeps session IDs stable across errors, and ignores late responses after session sweep.
  - Applied only to `openai-subscription` (chat and native responses). `toResponsesRequest` adds `forwardPromptCacheKey`; enabled for `openai-subscription` and `openai-responses` only. Affinity headers are authoritative and override caller `extraHeaders`.

<sup>Written for commit e6e6e650e4c16cb3d45ea7e7570f218242d7054a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2220?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/codex-subscription-cache-affinity.md
+++ b/.changeset/codex-subscription-cache-affinity.md
@@ -1,0 +1,5 @@
+---
+"manifest": patch
+---
+
+Restore prompt-cache hits on the ChatGPT subscription backend: send the session affinity headers the Codex CLI sends (`session-id`/`thread-id`, `x-codex-turn-state` replay, stable `prompt_cache_key`), and forward the caller's `prompt_cache_key` on OpenAI /responses conversions
--- a/packages/backend/src/routing/proxy/__tests__/chatgpt-request.spec.ts
+++ b/packages/backend/src/routing/proxy/__tests__/chatgpt-request.spec.ts
@@ -191,4 +191,36 @@ describe('ChatGPT Adapter – toResponsesRequest', () => {
     expect(result.instructions).toBe('You are a helpful assistant.');
     expect(result.input).toEqual([{ role: 'user', content: [{ type: 'input_text', text: 'Hi' }] }]);
   });
+
+  describe('prompt_cache_key forwarding', () => {
+    const body = {
+      messages: [{ role: 'user', content: 'Hi' }],
+      prompt_cache_key: 'conv-1',
+    };
+
+    it('forwards prompt_cache_key when opted in', () => {
+      const result = toResponsesRequest(body, 'gpt-5', { forwardPromptCacheKey: true });
+
+      expect(result.prompt_cache_key).toBe('conv-1');
+    });
+
+    it('drops prompt_cache_key by default', () => {
+      const result = toResponsesRequest(body, 'gpt-5');
+
+      expect(result).not.toHaveProperty('prompt_cache_key');
+    });
+
+    it.each([[undefined], [''], [42]])(
+      'ignores a %p prompt_cache_key even when opted in',
+      (key) => {
+        const result = toResponsesRequest(
+          { messages: body.messages, prompt_cache_key: key },
+          'gpt-5',
+          { forwardPromptCacheKey: true },
+        );
+
+        expect(result).not.toHaveProperty('prompt_cache_key');
+      },
+    );
+  });
 });
--- a/packages/backend/src/routing/proxy/__tests__/codex-session-affinity.spec.ts
+++ b/packages/backend/src/routing/proxy/__tests__/codex-session-affinity.spec.ts
@@ -70,6 +70,33 @@ describe('CodexSessionAffinity', () => {
       expect(headers).not.toHaveProperty('x-codex-turn-state');
     });
 
+    it('treats an over-long prompt_cache_key as absent to cap per-key memory', () => {
+      const longKey = 'x'.repeat(513);
+      const body: Record<string, unknown> = { prompt_cache_key: longKey };
+      const otherLongKey: Record<string, unknown> = { prompt_cache_key: 'y'.repeat(600) };
+
+      const first = affinity.prepare('token', body);
+      const second = affinity.prepare('token', otherLongKey);
+
+      // The giant key is never echoed back to the upstream…
+      expect(body.prompt_cache_key).toMatch(UUID_RE);
+      expect(body.prompt_cache_key).not.toBe(longKey);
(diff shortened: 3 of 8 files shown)


guillaumegay13:
Resolved the two CodeQL `js/insufficient-password-hash` alerts in cb0791f: the first version derived session ids by SHA-256-hashing the subscription token, which CodeQL rightly flags as a credential flowing into a fast unsalted hash. Since affinity ids only need to outlive the upstream prompt cache they pin (~minutes), deterministic derivation bought nothing — ids are now `randomUUID()`s memoized per session in the existing TTL map, so the credential never flows into any hash or outgoing header. Coverage on the file stays at 100% lines/branches.

PR #2242, opened 2026-06-12
feat: add AWS Bedrock provider

guillaumegay13:
### ✨ What changed
- Adds AWS Bedrock as an API-key provider through Bedrock Mantle.
- Adds Mantle region selection and short-term key region inference.
- Wires Bedrock model discovery and proxy routing to regional Mantle hosts.
- Adds frontend provider metadata, icon, docs link, tests, and smoke script.

### 💭 Why
Manifest needs Bedrock support without AWS SDK credentials in the app path.
Mantle matches Manifest's OpenAI-compatible provider flow.

### 👤 For users
Users can connect Bedrock with a Bedrock API key and route to discovered models.
Unsupported Mantle regions, including `eu-west-3`, are not offered.

### 📝 Notes
Live smoke passed against `us-east-1` with `mistral.ministral-3-8b-instruct`.
Cubic reports zero open issues and GitHub checks are green.

### DIFF
--- a/.changeset/bedrock-provider.md
+++ b/.changeset/bedrock-provider.md
@@ -1,0 +1,5 @@
+---
+'manifest': patch
+---
+
+Add AWS Bedrock as an API-key provider using Bedrock Mantle's OpenAI-compatible endpoints. Bedrock vendor-prefixed model IDs now resolve model parameters, pricing, and capabilities through the underlying provider metadata while keeping the original Bedrock ID for inference.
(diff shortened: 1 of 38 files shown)


guillaumegay13:
Closing after local validation. The branch boots locally: backend health is green on 127.0.0.1:3001, frontend renders on 127.0.0.1:3000, login works with the seeded dev user, and the Bedrock provider form renders without browser console errors.

It is not mergeable as-is:
- Frontend typecheck reproduces CI: `npx tsc --noEmit` fails at `packages/frontend/src/pages/Playground.tsx:352` and `:363` because `AvailableModel.display_name` is now `string | null | undefined`, but `findDisplayName` expects `string | undefined`.
- Backend tests reproduce CI: `npm test --workspace packages/backend -- --runInBand` fails `packages/backend/src/model-discovery/model-fallback.resolution-order.spec.ts` because `lookupWithVariants` now checks `anthropic/anthropic-claude-sonnet-4-6` before the expected dot variant `anthropic/claude-sonnet-4.6`.

Frontend Vitest itself passed locally: 181 files, 3755 tests.

Issue #2274, opened 2026-06-16
bug: subscription global providers page is very slow


guillaumegay13:
## Summary

The providers / global subscriptions page blocks on `GET /api/v1/providers`, whose **~7.2 s is TTFB** (time to first byte) — pure server processing *before* the response body downloads, so it's not network/bandwidth. The endpoint mixes two jobs: provider **config** (cheap) and provider **usage analytics** (expensive), running two sequential aggregations over `agent_messages` (3.7M rows / 8 GB) on every load.

```mermaid
flowchart LR
    subgraph BEFORE["Before · 7.2s"]
      A0["GET /api/v1/providers"] --> A1["config · ~50ms"]
      A0 --> A2["agent_messages 30d scan · ~1.0s"]
      A0 --> A3["agent_messages 7d scan · ~6.7s"]
    end
    subgraph AFTER["After · paints ~50ms, live"]
      B0["GET /api/v1/providers"] --> B1["config only · ~50ms"]
      C0["GET /api/v1/providers/usage"] --> C1["1 query + covering index · index-only"]
      S["SSE message / routing ping"] --> C0
    end
```

## Root cause

`TenantProvidersController.listProviders` decorates each provider with usage stats via two back-to-back queries over the hot telemetry table:

- 30-day consumption (tokens / cost / messages, last used) — ~1.0 s
- 7-day token sparkline — ~6.7 s

Confirmed on prod with `EXPLAIN (ANALYZE, BUFFERS)` — these are query times measured *inside Postgres*, network-independent: the `(tenant_id, timestamp)` index finds the rows in ~13 ms, but the **random heap fetches** for the non-indexed columns (cold cache, 8 GB table) are the 1–6 s. Two such scans, sequential, uncached, on every page load. The page renders nothing until it returns; `GlobalOverview` hits the same endpoint.

## Fix (in progress)

Separate config from analytics, and make the analytics cheap:

| Change | Effect |
| --- | --- |
| `GET /api/v1/providers` → config only (no `agent_messages`) | Page paints in ~50 ms |
| New `GET /api/v1/providers/usage` (analytics module): one collapsed query, UTC day buckets | Usage loaded separately, streamed into cells |
| Covering index `(tenant_id, timestamp) INCLUDE (provider, auth_type, input_tokens, output_tokens, cost_usd)` | Index-only scan, ~tens of ms even cold |
| Frontend: two resources; usage sourced on SSE `messagePing` / `routingPing` | Stats update live, no reload |

[...]

PR #2277, opened 2026-06-16
perf(frontend): SWR cache so dashboard navigation stops refetching everything

guillaumegay13:
Related to #2274 (the "every tab switch refetches everything" follow-up noted there).

## Problem
Every dashboard page fetches with raw `createResource`, and the router unmounts/remounts pages on navigation, so switching Overview → Messages → Subscriptions → back refetches **everything** every time. No cross-navigation cache.

## Change
A transparent stale-while-revalidate cache behind the shared `fetchJson` GET helper — pages need no edits.
- 30s TTL, stale-while-revalidate, in-flight dedupe; only resolved 2xx GETs cached.
- Deny-list for always-fresh/sensitive routes (`/auth`, `/key`, `/rotate-key`, `/events`, `/health`) + per-call `{ cache: false }` opt-out; mutations invalidate on success.
- **SSE-driven freshness**: each `message`/`agent`/`routing` event invalidates its key group **before** bumping the ping signal, so the ping-triggered refetch reads fresh data. Live updates are preserved — navigation is instant *and* never stale past the next SSE tick.

Compatible with #2274's `/api/v1/providers/usage` split: the `message` group (`/usage`) and `routing` group (`/providers`) substring-match that endpoint, so new-message live refresh of the usage widget keeps working once both land.

## Tests
Frontend 3896 passed, 100% line coverage on new/changed files (cache hit/miss/SWR/dedupe/TTL/deny-list/opt-out/mutation + per-ping invalidation ordering). Build + lint green.

---
## Summary by cubic
Adds a transparent SWR cache behind `fetchJson` GETs so dashboard tab switches stop refetching everything. Live data stays fresh via SSE-driven invalidation; addresses the follow-up in #2274.

- **New Features**
  - 30s TTL SWR cache for cacheable GETs with in-flight dedupe; caches only 2xx and keys by full URL.
  - Safety: deny-list (`/auth`, `/key`, `/rotate-key`, `/events`, `/health`); per-call `{ cache: false }`; `getModelParamSpecs` and `getCustomProviders` bypass shared SWR; mutations drop all GETs on success.
  - SSE invalidates URL groups (`message`, `agent`, `routing`) before ping bumps so refetches read fresh data; routing also clears the custom-providers cache.

- **Bug Fixes**
  - Prevent cache resurrection by suppressing writes from requests that started before an invalidation (generation guard), and scope error cleanup to the correct in-flight slot.
  - Prune fully expired, idle entries on successful writes to bound growth; ensure GlobalOverview usage/overview charts refetch on message pings by keying resources on `messagePing()`.

<sup>Written for commit 1f3b7901f28575f797899b5e29c91d038fbc5b9a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2277?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/frontend/src/pages/GlobalOverview.tsx
+++ b/packages/frontend/src/pages/GlobalOverview.tsx
@@ -180,12 +180,12 @@ const GlobalOverview: Component = () => {
 
   // ── Data resources (5 parallel) ──────────────────────────────────────
   const [overview] = createResource(
-    () => chartRange(),
-    (range) => getOverview(range) as Promise<OverviewResponse>,
+    () => ({ range: chartRange(), _ping: messagePing() }),
+    (p) => getOverview(p.range) as Promise<OverviewResponse>,
   );
 
   const [agents] = createResource(
-    () => agentPing(),
+    () => ({ _agentPing: agentPing(), _messagePing: messagePing() }),
     async () => {
       try {
         const data = await getAgents();
@@ -251,12 +251,12 @@ const GlobalOverview: Component = () => {
   };
 
   const [agentTimeseries] = createResource(
-    () => ({ range: chartRange(), group: groupBy() }),
+    () => ({ range: chartRange(), group: groupBy(), _ping: messagePing() }),
     (p) => tokenFetcher(p.range, p.group),
   );
 
   const [agentMessageTimeseries] = createResource(
-    () => ({ range: chartRange(), group: groupBy() }),
+    () => ({ range: chartRange(), group: groupBy(), _ping: messagePing() }),
     (p) => msgFetcher(p.range, p.group),
   );
 
@@ -266,7 +266,7 @@ const GlobalOverview: Component = () => {
   };
 
   const [agentCostTimeseries] = createResource(
-    () => ({ range: chartRange(), group: groupBy() }),
+    () => ({ range: chartRange(), group: groupBy(), _ping: messagePing() }),
     (p) => costFetcher(p.range, p.group),
   );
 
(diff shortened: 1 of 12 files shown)


guillaumegay13:
Both cubic findings addressed in f34477928 (frontend `services/api/cache.ts` + tests).

**P1 — in-flight response resurrecting an invalidated key (identified by cubic):** Added a module-level invalidation `generation` counter that every `invalidate*` bumps. A `cachedFetch` captures the generation at start and, on success, suppresses its cache write if the generation advanced while the fetch was in flight — so a response that started *before* an invalidation can no longer repopulate the just-dropped key. The post-invalidation refetch captures the new generation and writes normally, keeping the SSE invalidate-before-bump contract sound. Also scoped the `.catch` handler to the in-flight slot it actually registered (`current.inflight === inflight`) so a late failure never clobbers a newer request that took over the key.

**P2 — unbounded cache growth (identified by cubic):** Added `pruneExpired` that sweeps fully-expired, idle entries (no live `inflight`) on every successful write, so growth is bounded incrementally without a timer.

Coverage stays at 100% on `cache.ts` (3 new tests for the resurrection guards + 1 for pruning).

Issue #1930, opened 2026-05-17
feat(openrouter): improve generation log attribution (user, User-Agent, categories)


adaofeliz:
## Summary

When Manifest routes requests through OpenRouter, several fields in OpenRouter's generation logs are missing or uninformative:

| Field | Current value | Expected |
|---|---|---|
| `session_id` | `null` | end-user identifier |
| `external_user` | `null` | end-user identifier |
| `user_agent` | `"node"` | `"Manifest/1.0"` |

### `session_id` / `external_user`

OpenRouter populates both fields from the `user` field in the **request body** (a stable per-end-user identifier — see [OR API docs](https://openrouter.ai/docs/api/reference/overview#requests)). Manifest never injects this field into the forwarded body. Although OpenRouter is in `PASSTHROUGH_PROVIDERS` (so a `user` field _would_ pass through if present), no Manifest component sources it from the incoming request context.

The incoming request already carries the end-user identity via the `x-user-id` header (set by OpenWebUI and other agents via `ENABLE_FORWARD_USER_INFO_HEADERS`). Manifest needs to inject `user: <x-user-id>` into the forwarded body when the resolved provider is `openrouter`.

### `user_agent: "node"`

The `openrouter` endpoint in `provider-endpoints.ts` does not set a `User-Agent` header. Node.js `fetch` fills in `"node"` by default.

### 1. `provider-endpoints.ts` — static header additions (trivial)

```ts
openrouter: {
  buildHeaders: (apiKey: string) => ({
    Authorization: `Bearer ${apiKey}`,
    'Content-Type': 'application/json',
    'HTTP-Referer': 'https://manifest.build',
    'X-OpenRouter-Title': 'Manifest',       // preferred over X-Title per current OR docs
    'X-OpenRouter-Categories': 'cloud-agent',
    'User-Agent': 'Manifest/1.0',           // fixes user_agent: "node"
  }),
```

### 2. `proxy-types.ts` + `proxy-fallback.service.ts` + `provider-client.ts` — inject `user` into body

- Add optional `endUserId?: string` to `ForwardOptions`
- Thread `userId` from `ProxyRequestOptions` → `tryForwardToProvider` opts
- In `buildRequest` at the OpenAI-compatible path, inject `requestBody.user = ctx.endUserId` when `endpointKey === 'openrouter'` and the value is present

This keeps the injection OpenRouter-scoped and avoids leaking end-user identifiers to other providers.

## Impact

[...]

PR #2324, opened 2026-06-23
feat(openrouter): forward x-session-key as x-session-id for sticky routing and prompt cache

adaofeliz:
## Summary

When Manifest routes requests through OpenRouter, the `x-session-key` header (forwarded by agents like OpenWebUI via `FORWARD_SESSION_INFO_HEADER_CHAT_ID=X-Session-Key`) is consumed internally but never passed upstream.

OpenRouter supports an [`x-session-id` header](https://openrouter.ai/docs/guides/best-practices/prompt-caching) (equivalent to `session_id` in the request body) that provides two concrete benefits:

1. **Sticky routing** — pins all requests in the same conversation to the same provider, maximising prompt cache hits
2. **Observability grouping** — groups requests as a session in the OpenRouter console

## Root cause

`provider-endpoints.ts` builds OpenRouter's upstream headers without `x-session-id`. The `sessionKey` is already threaded through `ForwardOptions` → `proxy-fallback.service.ts` → `buildProviderExtraHeaders`, so the plumbing is in place — the `xai` entry in `PROVIDER_EXTRA_HEADER_BUILDERS` already uses the exact same pattern for `x-grok-conv-id`.

## Change

Added a single entry to `PROVIDER_EXTRA_HEADER_BUILDERS` in `provider-hooks.ts`, mirroring the existing `xai` pattern:

```ts
// before
xai: (sessionKey) => ({ 'x-grok-conv-id': sessionKey }),

// after
xai: (sessionKey) => ({ 'x-grok-conv-id': sessionKey }),
openrouter: (sessionKey) => ({ 'x-session-id': sessionKey }),
```

## Impact

- Prompt cache hits increase across a conversation when routing through OpenRouter → **direct cost reduction**
- Sessions become visible and grouped in the OpenRouter console
- Zero risk: `x-session-id` is ignored by OpenRouter if no prior context exists for the session

## Testing

Added 4 tests to `provider-endpoints.spec.ts` covering:
- `openrouter` returns `x-session-id` with the session key value
- Case-insensitive provider name match
- `xai` still works as before
- Providers with no extra headers return `undefined`

Closes / related to #1930

### DIFF
--- a/packages/backend/src/routing/proxy/__tests__/provider-endpoints.spec.ts
+++ b/packages/backend/src/routing/proxy/__tests__/provider-endpoints.spec.ts
@@ -5,7 +5,7 @@ import {
   resolveEndpointKey,
   PROVIDER_ENDPOINTS,
 } from '../provider-endpoints';
-import { resolveSubscriptionEndpointKey } from '../provider-hooks';
+import { buildProviderExtraHeaders, resolveSubscriptionEndpointKey } from '../provider-hooks';
 
 describe('buildCustomEndpoint', () => {
   it('strips trailing /v1 from base URL to avoid double /v1', () => {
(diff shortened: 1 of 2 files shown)


guillaumegay13:
thanks for contributing @adaofeliz !

adaofeliz:
thank you @guillaumegay13 !!! Love the project!

guillaumegay13:
@adaofeliz can you contact me please ? guillaume@manifest.build

PR #2325, opened 2026-06-24
fix: strip client-supplied MPS params unsupported by the resolved model

flaviohcardoso:
When a client sends a parameter like `temperature` that is a known MPS path but not listed in the resolved model's specs (e.g. claude-opus-4-7 which removed sampling controls), the proxy now strips it before forwarding to the provider. Previously it passed through unchanged, causing provider errors like:

  `temperature` may only be set to 1 when thinking is enabled

The fix adds `getAllKnownParamRootPaths()` to ProviderParamSpecService and uses it in ProxyFallbackService.applyParamMerge to compare the model's supported roots against the full MPS catalog, removing any client-supplied params that are known MPS paths but absent from the model's own specs.

---
## Summary by cubic
Strip unsupported MPS params from client requests to prevent provider errors (e.g., `temperature` on models without sampling). Only strip when the resolved model has specs; non-MPS fields and all params for uncataloged models are preserved.

- **Bug Fixes**
  - Added `getAllKnownParamRootPaths()` in `ProviderParamSpecService` to collect all known MPS root paths.
  - Updated `ProxyFallbackService.applyParamMerge` to drop known MPS params not in the resolved model’s specs, and to preserve all params when the model has no specs.

<sup>Written for commit 9640e3497f32cb16c400bf0d3d3a095b7c5ce447. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2325?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/backend/src/routing/proxy/__tests__/proxy-fallback.service.spec.ts
+++ b/packages/backend/src/routing/proxy/__tests__/proxy-fallback.service.spec.ts
@@ -578,6 +578,40 @@ describe('ProxyFallbackService', () => {
       expect(forwarded.body.messages).toBeDefined();
     });
 
+    it('preserves known MPS params when the resolved model has no specs', async () => {
+      providerClient.forward.mockResolvedValue({
+        response: new Response('{}', { status: 200 }),
+        isGoogle: false,
+        isAnthropic: false,
+        isChatGpt: false,
+      });
+      modelParamsService.get.mockResolvedValueOnce(null);
+      providerParamSpecs.getAllKnownParamRootPaths.mockReturnValue(
+        new Set(['temperature', 'top_p', 'max_tokens']),
+      );
+
+      await service.tryForwardToProvider({
+        provider: 'deepseek',
+        apiKey: 'sk-test',
+        model: 'new-provider-model',
+        body: {
+          messages: [{ role: 'user', content: 'hi' }],
+          temperature: 0.7,
+          top_p: 0.9,
+          max_tokens: 2048,
+        },
+        stream: false,
+        sessionKey: 'sess-1',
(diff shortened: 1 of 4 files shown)


guillaumegay13:
hi @flaviohcardoso, which issue does this solve ?
can't you configure the model parameters from Manifest UI instead of passing them in the SDK?

flaviohcardoso:
hey @guillaumegay13 this PR fixes a bug where parameters like `temperature` sent via the SDK (or any other source, in mycase vscode) could cause provider errros when the resolved model doesn't support them. For example, claude-opus-4-7 and 4-8 removed sampling controls, and sending temperature caused the Anthropic API to return an error.

This fix automatically strips MPS parameters that are unknown to the specific model, ensuring compatibility for those using the SDK for parameter overriding. The Manifest UI configuration continues to work as usual.  this logic simply ensures that invalid parameters are removed before reaching the provider to prevent API errors.

guillaumegay13:
Hi @flaviohcardoso, Manifest SDK being passthrough with the parameters is a choice we've made. Some people want to override Manifest parameters' configuration in the SDK. That being said, we'll discuss it internally regarding your use case. Keeping the PR open for now! Thanks for your explanation.

Feel free also to join the discord, it's a good way to discuss issues like this!

PR #2354, opened 2026-06-29
feat: add Codex CLI as a coding-assistant agent platform

guillaumegay13:
### ✨ What changed
- Register `codex` as a coding agent platform (picker tile, label, icon).
- Add a `~/.codex/config.toml` setup panel: Responses API (`wire_api = "responses"`), `mnfst_` key via `env_key`.
- On the chat/completions path, fold Responses `role: developer` into `system`.
- On the chat/completions path, drop hosted tools (`web_search`, `file_search`, ...); chat upstreams only accept `function` tools.

### 💭 Why
Codex CLI pointed at Manifest broke on non-OpenAI upstreams. The `developer` role and hosted tools both 400 chat/completions providers. The strict Responses SSE lifecycle Codex needs already ships (#2066, #2231), so only the platform plus these two conversions were missing.

### 👤 For users
Codex shows up in the agent picker with a copy-ready config. A Codex CLI on a self-hosted instance completes requests without retries.

### 📝 Notes
- Native Responses path (OpenAI) keeps `developer` roles and hosted tools intact.
- Draft: open on changeset bump level, whether to add a Wingman profile, and swapping the placeholder icon for an official mark.


Wiktor102:
It seems like Codex has changed how profiles are handled:

> Error loading config.toml: --profile `manifest` cannot be used while C:\Users\wikto\.codex\config.toml contains legacy `profile = "manifest"` or `[profiles.manifest]` config; move those settings into C:\Users\wikto\.codex\manifest.config.toml and remove the legacy profile selector/table. See https://developers.openai.com/codex/config-advanced#profiles for more information.

I think this is actually a good change since users could download a ready-to-go profile config from the manifest dashboard.

guillaumegay13:
Hi @Wiktor102, can you provide more details please? Do you need to use Manifest in Codex ?

I checked & it seems we can only set it up for Codex CLI, not the Codex app so it feels limited to me.

Wiktor102:
I think this was almost working in the Codex app like a week ago, but right now it is indeed blocked.
- With ChatGPT account:
image

- With an api key account:
image

The CLI works, but you have to set up the profile as a separate `.config.toml` file. Otherwise, it won't launch with the error I posted earlier. So the onboarding should be changed to reflect that. I did, however, encounter [this comment](https://github.com/openai/codex/issues/22160#issuecomment-4422466742) from an OpenAI employee saying they will be deprecating profiles.

I also found a mention of the `model_catalog_json` option in their docs.  It would be fun if this allowed for direct model selection, but I haven't played with it yet. [Here](https://platform.minimax.io/docs/token-plan/codex#configure-model-capabilities-optional) is some reference of that JSON schema I managed to find on Minimax's website - interestingly enough, they say the desktop app **is** supported.

PR #2478, opened 2026-07-11
fix: preserve Claude and Codex gateway fidelity

ziahamza:
## Why

Manifest already accepts Anthropic Messages and OpenAI Responses traffic, but coding clients depend on more than the JSON schema: they send client/session metadata, beta capabilities, native tool types, and streaming event lifecycles that must survive the gateway boundary. Reconstructing or leaking these fields across providers can break Claude Code or Codex, especially during fallback.

## What changed

- Detect Claude Code and Codex request context, bound header sizes, hash derived session affinity, and forward only safe protocol metadata to the matching native subscription upstream. Caller credentials, cookies, and account headers never cross the boundary; provider-owned auth/account metadata remains authoritative.
- Keep native Claude Messages bodies intact on Anthropic routes, preserve live beta/client headers, pass through valid Anthropic errors and safe response metadata, and return Anthropic-shaped Manifest errors on `/v1/messages`.
- Preserve native Codex Responses bodies/headers on OpenAI subscription routes, retain account/FedRAMP metadata from OAuth ID-token claims without storing the ID token, and return OpenAI-shaped Manifest errors on `/v1/responses`.
- Upgrade cross-provider fallback translation for mid-conversation system messages, reasoning effort, structured output, function/custom tools, parallel fragmented tool streams, completed output reconstruction, and terminal `response.failed` events.
- Keep Anthropic client-executed tools available across fallback while removing Anthropic server tools; omit Responses-hosted tools and any orphaned `tool_choice` that generic Chat Completions providers cannot execute.
- Thread protocol context and subscription metadata through primary, auto-fix, OAuth-refresh, and fallback paths, including model discovery.
- Close the final review gaps: provider-generated credentials/account routing override Playground headers, required Claude capability headers survive header budgets, native Claude OAuth requests receive their identity/cache block without rewriting other fields, Playground preserves OpenAI workspace metadata across refresh, split OAuth claims are merged, Anthropic errors share one status map, and conflicting credentials return actionable guidance.

## User impact

- Claude Code can use Manifest as `ANTHROPIC_BASE_URL` without losing native request semantics when Anthropic is selected, and can receive a valid Anthropic stream when Manifest falls back to Codex/OpenAI.
- Codex can use Manifest as its Responses endpoint without losing native request semantics when the OpenAI subscription route is selected, and can receive a valid Responses stream when Manifest falls back to Anthropic.
- Generic Chat Completions behavior remains unchanged, including the existing friendly setup responses.

## Validation

- `npm run build --workspace=packages/shared`
- `npm run build --workspace=packages/backend`
- ESLint on all changed backend source/tests — 0 errors
- `cd packages/backend && npx jest --coverage --runInBand` — 415 suites, 7,697 tests passed
- `npm run check:changesets`
- Prettier check and `git diff --check`
- Disposable real-client E2E with isolated config and mock upstreams:
  - Claude Code 2.1.207: native Anthropic request returned 503, Manifest fell back to the OpenAI subscription Responses route, and Claude completed with `PONG`. Claude headers/body reached only Anthropic; no Claude metadata leaked to OpenAI.
  - Codex 0.144.1: native OpenAI subscription request returned 503, Manifest fell back to Anthropic Messages, and Codex completed with `PONG`. Codex headers/body reached only OpenAI; no Codex metadata leaked to Anthropic.

## Notes

- This is backend-only and independent of #2354's UI/config work.
- Native routes preserve protocol fidelity. Cross-provider translation is necessarily best-effort: provider-hosted/server-only tools and context-management features with no safe analogue are omitted instead of being sent as invalid or caller-executed tools.
- This changes transport compatibility and credential isolation; it does not alter provider subscription terms or confer authorization to proxy a consumer subscription.

---
## Summary by cubic
Preserves Claude Code and Codex protocol fidelity by forwarding only safe, endpoint-aware metadata and passing through native streams, headers, and errors. Adds robust streaming with heartbeats and strict deadlines, OpenAI workspace-aware OAuth routing, a local token counter, and a content-blind canary; keeps Claude Permission Auto on Anthropic and honors the `default` classifier model.

- **New Features**
  - Endpoint-aware header forwarding; block secret/provider headers; reapply provider auth only on native routes.
  - Native passthrough for Anthropic Messages and OpenAI Responses, including Responses `custom` tools; map JSON schema and `reasoning_effort`; stream Codex plan reasoning; qualify Google streaming (incl. CodeAssist envelope).
  - Streaming and budgets: SSE heartbeats with `PROVIDER_STREAM_TIMEOUT_MS`, 105s pre‑response deadline, attempt‑scoped semantic output timeout; accept chunked request bodies; local `/v1/messages/count_tokens`.
  - OAuth/discovery: extract and persist OpenAI `accountId`/FedRAMP for workspace routing and store in OAuth blob metadata; send Codex `version` and updated `user-agent`; preserve OpenAI workspace metadata across refresh and model discovery.
  - Claude: keep Permission Auto classifier calls on Anthropic and resolve the `default` classifier alias; normalize Opus 4.8 sampling; update setup snippet to use `model: "default"`.
  - Content‑blind canary with trace IDs (`npm run canary:gateway`).

- **Bug Fixes**
  - Prevent cross‑provider metadata leakage; surface actionable errors on mixed credentials; pass through native protocol error bodies for Messages/Responses.
  - Stream integrity and recovery: reject empty/malformed/truncated SSE and tool args; terminate stalled streams; keep quiet streams alive with heartbeats; recover terminal text when deltas are omitted; treat max‑output exhaustion without output as retryable before streaming (incl. Google/CodeAssist); retry a silent Codex primary once within budget.
  - Timeouts/sampling: enforce pre‑response vs. streaming budgets; preserve provider timeout env fallback.
  - Security/telemetry: key agent‑token cache digests with a process‑local HKDF salt; add per‑request trace IDs; finalize canceled client requests with proper telemetry; preserve canceled attempt ordering; close all pending upstream attempts on client disconnect and record a 499 outcome.
  - CI: ensure chunked invalid‑key probes return 401; retry the gateway canary on transient Manifest concurrency.

<sup>Written for commit fc3c7b22c1d7e2da5e7570788d9f56ea654663df. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2478?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## July 18 production hardening

Follow-up production reports exposed two more gateway failure modes: upstream HTTP 200 responses with empty/malformed bodies or truncated SSE streams were being recorded as successful completions, and Claude Code’s optional `/v1/messages/count_tokens` probe returned 404, causing an extra Haiku inference that amplified M203 concurrency failures.

This update:

- requires meaningful text/reasoning or a named tool call plus a real terminal event before recording success;
- emits native Anthropic error events and failed telemetry for empty/truncated streams, including native Anthropic passthrough;
- rejects empty/malformed non-streaming HTTP 200 responses;
- serves authenticated local token-count estimates without consuming an inference slot;
- adds HTTP E2E coverage and a Docker smoke regression requiring a chunked invalid-key probe to return 401, never 411;
- parses classifier beta flags as comma-delimited tokens, addressing the latest Cubic P2 review note.

Additional validation:

- backend unit: 418 suites, 7,752 tests passed;
- PostgreSQL E2E: 41 suites, 349 tests passed;
- shared/backend builds, full lint, changeset and diff checks passed.

## July 20 content-blind contract monitoring

The latest production follow-up adds content-blind diagnostics and an exact-output canary without retaining or printing model output:

- malformed SSE payloads and malformed tool arguments now terminate as native Anthropic error events and failed telemetry;
- every gateway request receives an x-manifest-trace-id that is also stored with telemetry, so client reports can be correlated without prompts or response bodies;
- stream logs report only event, block, character, tool, malformed-event, terminal-event, and stop-reason metadata;
- premature client disconnects emit a sanitized warning with trace, API mode, provider, and model;
- npm run canary:gateway performs a real streaming exact-text contract check but outputs only status, timing, trace, counts, model metadata, and SHA-256 digests.

Additional validation:

- backend unit: 418 suites, 7,785 tests passed;
- script unit: 28 tests passed;
- full repository build, backend lint, Prettier, changeset, and diff checks passed.


ziahamza:
Addressed the latest Cubic P2: classifier detection now tokenizes comma-delimited `anthropic-beta` flags and requires a flag that starts with `claude-code-`; a substring such as `future-claude-code-compatibility-*` follows normal routing. Added a regression test.

ziahamza:
Final July 18 production validation for head `0b3fec423`:

- All PR checks are green, including PostgreSQL backend, frontend, lint, CodeQL, Codecov patch/project, amd64/arm64 image validation, Compose smoke, install E2E, and Cubic.
- Public HTTP/1.1 chunked invalid-key probe returns 401 (never 411).
- Authenticated `/v1/messages/count_tokens?beta=true` returns a valid local estimate without an inference row.
- Claude Code 2.1.212 and Codex CLI 0.144.6 both completed real public-gateway turns through the OpenAI subscription route.
- Classifier-signature requests emitted the required preservation log and forwarded to `https://api.anthropic.com/v1/messages`; native Anthropic non-stream and stream probes both completed, and the stream contained a real `message_stop`.
- Post-deploy telemetry: 2 successful OpenAI subscription rows and 2 successful Anthropic subscription rows; no 411, 524, empty/malformed response, premature EOF, or origin connection signals.
- One deliberate synthetic probe that included the 1M-context beta received Anthropic's expected 429 requiring usage credits; removing that unrelated beta produced both successful native validations.

Rollback remains available as `manifest-zia-release:pre-stream-integrity-20260718`.

ziahamza:
Production follow-up added in c8705748e: the provider timeout previously ended when streaming headers arrived, so body reads could continue for 120–227 seconds and hit Cloudflare 524 or surface as premature EOF. The fix keeps the 90-second deadline active through full body consumption, emits protocol-native terminal errors for timeout/reset, and records 504/502 failed telemetry rather than a false `ok`. Validation: backend build/type-check/lint passed; 418 suites and 7,782 tests passed, including partial OpenAI subscription output translated to Anthropic followed by a mid-stream timeout.

ziahamza:
Follow-up for the production Cloudflare failure window is now pushed in 211be8264.

After the full-body deadline fix was deployed, four real streaming turns that previously would have ended as silent/truncated HTTP 200s correctly surfaced as 504 `stream timed out` failures at the configured 90s deadline. There were no 411 regressions or cloudflared origin errors. This follow-up keeps streaming responses active through proxy-safe SSE comment heartbeats every 15s and introduces a separate bounded `PROVIDER_STREAM_TIMEOUT_MS` (default 180s), while buffered calls retain `PROVIDER_TIMEOUT_MS`. Heartbeats bypass content/terminal validation and therefore cannot turn an empty completion into success.

Validation: 418 backend suites / 7,783 tests passed, including the quiet-stream heartbeat regression; typecheck, build, lint, Compose config, and diff checks passed.

ziahamza:
Production semantic-canary follow-up (2026-07-20): the SOL route passed the exact-text and complete-SSE contract. The Opus 4.8 route reached Anthropic but returned HTTP 400 because `temperature` is deprecated for that model. Commit 1350b84fb omits the optional field from the canary and defensively removes it at the Anthropic provider boundary for both translated and native Claude Code requests. Focused verification: 378 backend tests, 6 canary tests, backend build, lint, formatting, and diff checks all pass.

PR #2490, opened 2026-07-13
feat(analytics): add auto_fixed count to errors/breakdown

guillaumegay13:
## ✨ What changed
- Added an `auto_fixed` field to the `GET /api/v1/errors/breakdown` response — the number of requests healed by Auto-fix in the selected window.
- Counted from the failed-original row (`status='auto_fixed'`), so it is exactly **one per healed request** (the paired `ok` retry is a separate row) — no double-counting.
- Added a typed `getErrorBreakdown(range, agentName?)` + `ErrorBreakdownResponse` to the frontend analytics API client so a dashboard can consume it in one import.
- No UI, no new endpoint, no DTO change — the existing dynamic `range` (1h/6h/24h/7d/30d/90d/365d), tenant scoping, and caching all carry over unchanged.

## 💭 Why
Auto-fix (self-healing) already stores healed requests as a linked pair of `agent_messages` rows, but nothing aggregates them for the dashboard. Surfacing the heal count next to the existing error breakdown lets someone build the Auto-fix dashboard on top without inventing a new query path.

Counting can't be done client-side: `autofix_applied` is `true` on **both** rows of a healed pair, so a naive `WHERE autofix_applied` doubles the number; and tenant isolation must stay server-side. The correct, deduplicated `COUNT(*) FILTER (status='auto_fixed')` belongs in the backend so every surface agrees.

**Important:** `auto_fixed` is **not additive** with `total_errors`. The healed original is classified with an `error_origin` and marked `superseded`, so it is already inside `total_errors`/`by_origin`. Read `auto_fixed` as "of those errors, this many were auto-fixed", never as a separate bucket to sum. This is documented on the field in both the backend and frontend types.

## 👤 For users
No visible change yet — this is the data foundation for an Auto-fix dashboard.

## 🔧 For operators
No operator impact. No migration, no config, no new route.

## 🧪 How to test
1. `cd packages/backend && npx jest src/analytics/services/error-breakdown.service.spec.ts` — 10 passing, 100% line coverage on the service.
2. `cd packages/frontend && npx vitest run tests/services/api/analytics.test.ts` — 15 passing (includes `getErrorBreakdown`).
3. Call `GET /api/v1/errors/breakdown?range=7d` and confirm the response now includes `auto_fixed`.

---
## Summary by cubic
Expose an `auto_fixed` count on `GET /api/v1/errors/breakdown` so dashboards can show how many errors were healed by Auto-fix in the selected window. Also adds a typed `getErrorBreakdown()` to the frontend for easy consumption.

- **New Features**
  - Added `auto_fixed` to the breakdown response; counted from the failed-original row (`status='auto_fixed'`) for exactly one per healed request; not additive with `total_errors`.
  - Reused existing range presets, tenant scoping, and caching; no new route.
  - Frontend: added `getErrorBreakdown(range, agentName?)` and `ErrorBreakdownResponse` in the `analytics` API client.

<sup>Written for commit 3a3c8d58b8e2a8d2ef32967978898b018fd9115c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2490?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/autofix-count-errors-breakdown.md
+++ b/.changeset/autofix-count-errors-breakdown.md
@@ -1,0 +1,5 @@
+---
+"manifest": patch
+---
+
+Add an `auto_fixed` count to the `GET /api/v1/errors/breakdown` response (number of requests healed by Auto-fix in the window), plus a typed `getErrorBreakdown()` frontend API wrapper — so a dashboard can surface auto-fixed alongside errors.
--- a/packages/backend/src/analytics/services/error-breakdown.service.spec.ts
+++ b/packages/backend/src/analytics/services/error-breakdown.service.spec.ts
@@ -8,7 +8,7 @@ interface GroupRow {
   count: string;
 }
 
-function makeQb(groups: GroupRow[], successful: number) {
+function makeQb(groups: GroupRow[], successful: number, autoFixed: number) {
   return {
     select: jest.fn().mockReturnThis(),
     addSelect: jest.fn().mockReturnThis(),
@@ -17,12 +17,16 @@ function makeQb(groups: GroupRow[], successful: number) {
     groupBy: jest.fn().mockReturnThis(),
     addGroupBy: jest.fn().mockReturnThis(),
     getRawMany: jest.fn().mockResolvedValue(groups),
-    getRawOne: jest.fn().mockResolvedValue({ count: String(successful) }),
+    // querySuccessful resolves first, queryAutoFixed second (Promise.all order).
+    getRawOne: jest
+      .fn()
+      .mockResolvedValueOnce({ count: String(successful) })
+      .mockResolvedValue({ count: String(autoFixed) }),
   };
 }
 
-function makeService(groups: GroupRow[], successful: number) {
-  const qb = makeQb(groups, successful);
+function makeService(groups: GroupRow[], successful: number, autoFixed = 0) {
+  const qb = makeQb(groups, successful, autoFixed);
   const repo = {
     createQueryBuilder: jest.fn().mockReturnValue(qb),
   } as unknown as Repository<AgentMessage>;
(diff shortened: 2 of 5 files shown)


SebConejo:
Reviewed this. It works, and merging it is safe for users who don't have Auto-fix: for them the field simply comes back as `0` and nothing else changes. A few things I noticed while going through it:

**1. The count could silently become wrong later.**
Today one healed request always leaves exactly one `auto_fixed` row, so counting rows gives the right number. But the recording side is deliberately written to allow *several* failed attempts per healed request. The day Auto-fix retries more than once, this counter starts reporting more heals than actually happened — and nothing will warn us, the number just quietly inflates. Counting healed requests (the linked groups) instead of rows would stay correct forever.

**2. The new field is never checked against a real database.**
The unit test mocks the query builder entirely, so what's actually being verified is the mock, not the SQL. The existing e2e spec for this endpoint wasn't extended. Concretely: if the status filter or the tenant scoping on the new query were wrong, every test would still be green.

**3. The test setup can't catch a missing tenant filter on this specific query.**
All three queries share a single mocked builder, so the assertion "the auto_fixed filter is applied" passes regardless of *which* query applied it. Removing the tenant scoping from just the new query wouldn't fail a single test. Combined with point 2, the tenant isolation of this count is effectively unverified.

**4. The response shape is now maintained by hand in two places.**
Backend and frontend each carry their own copy of the same interface, kept in sync by a comment. Next time someone adds a field on one side only, the two drift apart and nothing complains — the frontend cast doesn't validate anything at runtime.

**5. Pre-existing, but this PR makes it visible for the first time:**
healed originals still count as live provider errors in `provider_error_rate`. So an agent with Auto-fix ON shows a *worse* provider error rate per healed request than the same agent with it OFF — the opposite of what the feature does. Not caused here, but since a dashboard is about to surface these numbers side by side, probably worth a follow-up issue before that dashboard ships.

None of this blocks merging in my view — the code is correct as of today. 1 and 2 are the ones I'd address while the context is fresh.

guillaumegay13:
Closing as redundant. The redesigned dashboard PR #2485 already contains this exact `auto_fixed`-in-errors-breakdown change — the `ErrorBreakdownService.queryAutoFixed()` implementation, the `auto_fixed` field + doc comment, and the SQL are byte-for-byte identical there. Consolidating the auto-fix analytics into #2485 so there's a single owner for the errors-breakdown change; no code is lost. The branch stays if we ever need to cherry-pick it independently.

PR #2503, opened 2026-07-14
feat: separate requests from provider attempts

guillaumegay13:
### ✨ What changed

- Adds explicit request parents and provider-attempt children across recording, Playground, analytics, and the Requests UI, including zero-attempt rejections and rolled-up cost, token, and latency data.
- Keeps unlinked history visible as synthetic single-attempt requests and surfaces request success, provider-attempt success, recovered requests, and Manifest lift.

### 💭 Why

Provider hops were conflated with caller requests, which made fallback behavior, counts, and success rates ambiguous.

### 🔧 For operators

The deploy migration uses a short lock timeout, while history regroups after readiness in throttled, resumable batches. It keeps `agent_messages` as an updatable compatibility view for old replicas; remove it through #2507 only after this release is deployed and drained.

---
## Summary by cubic
Separates caller requests from provider attempts to fix counts and reliability metrics, adds attempt ordering and lifecycle tracking, and normalizes status to pending | success | failed across both layers. Updates analytics, proxy, and billing to work off request parents while preserving historical behavior during backfill.

- **New Features**
  - Data model: parent `requests`, child `provider_attempts` (includes zero‑attempt Manifest rejections); canonical status on both: `pending | success | failed`.
  - Attempt order & lifecycle: `attempt_number` per request; proxy records start/finish for each attempt and links them; locally rejected requests finish without creating an attempt; skips transport failures on client aborts.
  - Analytics: overview/timeseries counts use `requests`; provider analytics adds attempt‑reliability aggregates (defaults to zero when missing); “has data” treats zero‑attempt requests as data; readers accept legacy status aliases.
  - Proxy/Playground: generate a `requestId`, record `requests` and all linked attempts across fallbacks; dedup by `requestId` then `traceId`/session.
  - Auto‑fix: store request‑level `autofix_status`; persist decision/explanations on attempts; details API returns status and reasons.
  - Feedback: store on `requests`; specificity feedback accepts a request ID and resolves one specificity attempt.
  - Billing/usage: quotas count distinct `requests` with ≥1 attempt; provider usage scans `provider_attempts` only.

- **Migration**
  - `1801000000000-AddRequestsAndProviderAttempts`: creates `requests` (incl. `autofix_status`), renames storage to `provider_attempts`, adds nullable `request_id` with a CONCURRENT index, and creates a rollout‑safe `agent_messages` compatibility view.
[...]

PR #2504, opened 2026-07-15
feat: expose tenant Auto-fix access cohort for panel gating

guillaumegay13:
### ✨ What changed

Exposes the tenant's Auto-fix access cohort to the frontend so the redesigned dashboard can gate **only the Auto-fix-specific pieces** — not the whole overview.

- **Backend:** new tenant-level `GET /api/v1/autofix/cohort` → `{ eligible }`, backed by the existing `AutofixService.hasAccess` (the `AUTOFIX_ROLLOUT` phase + per-tenant `autofix_access_granted_at` / `autofix_waitlist_at`) — the same gate as the Auto-fix feature itself. Registered on `AutofixModule`.
- **Frontend:** `getAutofixCohort()` typed client (always-fresh, no SWR cache) + `AutofixCohort` type. No UI wiring here — #2485 consumes it to gate its auto-fixed KPIs/panels.

### 💭 Why

The request/attempt schema split from #2503 is **universal** — it applies to every tenant, and fallbacks happen to everyone. So the new request/attempt/fallback dashboard should ship to everyone; gating the whole overview by the Auto-fix cohort would mismatch the data for most users (a request-count KPI over an attempt-level list, etc.). The Auto-fix cohort is the wrong axis for the dashboard — it belongs only around the auto-fixed metrics, which are simply empty for a tenant outside the cohort.

This PR therefore ships just the eligibility signal. Membership is entirely backend-driven — no emails or user IDs in frontend code.

### ✅ Tests

- Backend controller spec: eligible, default, and null-tenant paths.
- Frontend service test: eligible + default.
- New files at 100% line coverage.

Verified end-to-end against a local cloud-mode backend: `401` unauthenticated, `{"eligible":false}` for a default tenant, `{"eligible":true}` once `autofix_access_granted_at` is set.

### 🧱 Stacking

Stacked on #2503 (`guillaumegay13/nonblocking-request-backfill`, hosted on mnfst) — this PR's base is that branch, so the diff is only the eligibility delta. Retarget to `main` once #2503 merges.

Aggregation endpoints for requests / attempts / fallbacked / auto-fixed counts follow as a separate stacked PR; the redesigned dashboard that consumes this cohort signal is #2485.


guillaumegay13:
Superseded: this landed on main via the rewritten request-first stack (#2485, #2534). Closing.

PR #2512, opened 2026-07-16
docs: define request and attempt terminology

guillaumegay13:
### ✨ What changed
- Makes `docs/glossary.md` the canonical Request and Provider Attempt contract.
- Defines shared statuses, recovery priority, and metric boundaries.
- Clarifies Provider Connection attribution and legacy status mappings.
- Keeps both agent guides as short links to the glossary.

### 💭 Why
The request and attempt implementation needs one agreed contract before schema and code changes.

### 📝 Notes
Contract only. No UI or runtime behavior changes.

Schema implementation follows in the request and attempt stack.

https://github.com/mnfst/manifest/pull/2503

https://github.com/mnfst/manifest/pull/2485


SebConejo:
Review of the glossary, internal consistency only (the doc is the source of truth, the code will align with it). The directional definitions, the examples table and the "applying a method is not enough" rule are excellent. Below are the gaps we found, as expression / current definition / problem:

| Expression | How the document defines it | Problem |
|---|---|---|
| **Last Attempt** | "The final provider attempt made for a Request. It is derived from the attempt chain; a zero-attempt Request has none." | "Final" is not defined: last by what? Time, a logical order, an index? The document does not say. |
| **Recovered by fallback** | "The successful Last Attempt has a non-null `fallback_from_model`." | Relies on an assumption the document never states: that a successful Request's Last Attempt is always the successful attempt. One sentence stating that invariant is missing. |
| **Recovered Requests** | "The union of the two groups, counted once per Request." | No priority rule when a Request satisfies both criteria (Auto-fix AND fallback). The UI has to place it in exactly one category. |
| **Successful Request** (used in "Request success rate") | Never defined. The formula says "Successful Requests divided by all Requests". | The central word of the metric has no definition, and the possible values of `requests.status` are not listed anywhere. |
| **Successful Attempt** (used in "Attempt success rate") | Never defined. The formula says "Successful Attempts divided by all Attempts". | Same problem, and the possible values of `provider_attempts.status` are not listed either. |
| **superseded** | Never defined. Cited once in the data-mapping table ("Routing and recovery context"). | A term cited in the canonical reference without a definition. |
| **Total Requests** | "Count `requests.id` once." | The perimeter is unstated: do we count in-flight Requests? Playground traffic? historical attempts with no parent Request? |
| **The 4 core definitions** (Request, Attempt, Last Attempt, Recovered Request) | Written out in full in 3 files: AGENTS.md, CLAUDE.md and the glossary. | Three copies of the same text: the first edit to a single file makes them drift. |

SebConejo:
One more point that directly affects the "Last Attempt" definition: attempt timestamps are currently fabricated, not measured.

The recorder writes nothing while a request is in flight. When the outcome is known, it writes every row in one batch, and since nobody captured the real time of each attempt along the way, it invents timestamps from a single base time plus 100ms offsets per row. In the fallback-exhausted flow it even dates the primary failure AFTER the fallback rows (`proxy-response-handler.ts`: `primaryTs = baseTime + (failedFallbacks.length + 1) * 100`), the reverse of what physically happened. So "last by timestamp" is display staging, not chronology.

This gives the "Last Attempt" definition two options, and the doc should pick one explicitly:

1. **Logical definition**: the Last Attempt is the attempt that CONCLUDED the Request (the successful one when it exists, else the last real failure). No dependency on timestamps; this is what the Requests list rendering already does.
2. **Chronological definition**: keep "final attempt in time", which then REQUIRES capturing real per-attempt start/end times in the forward path (the `FailedFallback` entries and the autofix context already travel through the flow; they just need to carry real times so the recorder stops inventing them). Bonus: per-attempt latency becomes measurable, and "something that happened at 12:10 is stored as 12:10".

Option 2 is the more honest data model long-term; option 1 works today without touching the recorder. Either way, the glossary should state which one "final" means.

PR #2519, opened 2026-07-17
feat: warn on reset-password page when no email provider is configured

### LINKED ISSUE #2516: `POST /api/auth/forgot-password` returns 404 and `/api/auth/reset-password` returns 400; self-service password change is unusable on self-hosted installs with `credential` provider type
## Summary

The forgot-password / reset-password flow is broken end-to-end on self-hosted installs using the `credential` provider type. There is no working path for a user to change or recover their own password without directly modifying the database, and no UI surface to do so. The two API endpoints that *should* implement this flow each fail in a different way, both pointing at the same root cause: a missing `forgotPassword` plugin path in the installed binary.

---

## Environment

- Self-hosted Manifest (single container, app-data on a persistent volume)
- Authentication provider: `credential` (username + password; not OAuth / SSO)
- Account is intentionally on `credential` provider type by design
- Manifest version: the build deployed at the reported install does not register the `forgotPassword` plugin route that the canonical UI documentation references

---

## Steps to Reproduce

1. Deploy Manifest self-hosted with the `credential` provider type configured.
2. Create or log in as a `credential` provider user.
3. From the login page (or `https://<host>/forgot-password` if linked), submit the user's registered email address via the documented JSON body.
4. Independently, attempt to call the `reset` endpoint directly without first obtaining a token from the `forgot` endpoint.

---

## Expected Behavior

`POST https://<host>/api/auth/forgot-password` with body `{"email": "<user>@<domain>"}`:

- Returns `200` (or `202`).
- Generates a one-time-use password-reset token.
- Persists the token (DB or kv store), hashed, with a finite expiry (15–60 minutes).
- Queues a transactional email containing a link of the form `https://<host>/reset-password/<token>`.
- Returns a constant-time response regardless of whether the email exists, to prevent account enumeration.

`POST https://<host>/api/auth/reset-password` with body `{"token": "<token>", "password": "<new-password>"}`:

- Validates the token (one-time use, not expired, matches the stored hash).
- Updates the stored credential hash using the same hashing algorithm that authenticates the user at sign-in.
- Invalidates the token.
- Returns `200`.

The web UI profile page should also expose a password-change form (or clearly document that changes must go through the email-reset path).

---

### Test 1 — `POST /api/auth/forgot-password`

```bash
$ curl -X POST https://<host>/api/auth/forgot-password \
       -H 'Content-Type: application/json' \
       -d '{"email": "<user>@<domain>"}'

HTTP/1.1 404 Not Found

{"message": "Not Found", "status": "error"}
```

The route is not registered by any installed plugin. Inspecting the running container confirms no plugin handler is loaded for that path.

### Test 2 — `POST /api/auth/reset-password` (no token / forged token / empty body)

```bash
$ curl -X POST https://<host>/api/auth/reset-password \
       -H 'Content-Type: application/json' \
       -d '{}'

HTTP/1.1 400 Bad Request

{"message": "token is required", "status": "error"}
```

Without a way to mint a valid token via `/forgot-password`, this endpoint can never succeed end-to-end. The 400 is the same root cause as the 404, viewed from the other end of the workflow.

### Test 3 — `/settings` or `/profile` UI

The web UI renders no "Change Password" form at all for `credential` provider accounts. There is no documentation linking this constraint to the absent forgot-password flow. Operators have no escape hatch within the application.

---

## Root Cause

From internal investigation:

1. The `forgotPassword` plugin functionality is not implemented in the installed Manifest binary. The route handler that *would* validate the email, generate a token, persist it, and queue an email is missing.
2. Even if the plugin were added, Manifest self-hosted has no SMTP / transactional-email transport wired in. Self-hosted Manifest does not support native SMTP; the canonical alternative is `Resend`, `Mailgun`, or `SendGrid` supplied via environment variables.
3. Reference documentation (sign-up confirmation copy and login UI text) implies the email-reset path exists, but the code path does not.

There are two missing pieces, not one. This issue documents both so they cannot be silently half-fixed.

---

## Workaround Currently in Use

Direct manual update of the credential user's password hash in the database, hashed with the algorithm Manifest expects for `credential` provider logins.

This is unsafe as a runbook:

- It bypasses any audit trail Manifest might (or might not) be expected to provide.
- It scales poorly once there is more than one operator or more than one user.
- It is brittle: the hashing algorithm is not externally documented and can change between Manifest versions.

If the only available recovery path is manual DB write, that itself is the bug being reported.

---

## Desired Solution

1. Implement the `forgotPassword` plugin such that `POST /api/auth/forgot-password`:

   - Returns `200` (or `202`) with a generic body that does not leak whether the email exists.
   - Generates a token, persists its hash (not the raw token), and queues an email using whichever transport is configured.

2. Implement `POST /api/auth/reset-password` such that with a valid token it:

   - Verifies the token against the persisted hash.
   - Rejects used, missing, or expired tokens with `400` and a generic error.
   - On success, updates the credential hash and invalidates the token.

3. Add a self-hosted email transport path, pluggable via environment variable (for example, `MANIFEST_EMAIL_PROVIDER=resend|mailgun|sendgrid|none`, plus provider-specific keys). With `none`, the plugin should still mint and persist the token (so admins can recover users via support) but log the email contents instead of sending.

4. Document the flow in the self-hosted deployment guide so operators can wire the transport and verify the flow end-to-end without inspecting the binary.

5. Update the web UI `/profile` or `/settings` page to surface a password-change form (which itself calls the forgot-password flow under the hood), so users have a self-service path that does not depend on email-loop timing.

---

## Security Considerations

- Token storage should be hashed, never stored in plaintext.
- Token single-use, with expiry of 15–60 minutes.
- Constant-time response on `/forgot-password` to prevent user enumeration.
- Rate limiting per email and per IP.
- All reset attempts should be logged, including the hashed token or request id, but never the raw token or password.

---

## Reproduction Artifact (anonymized)

```bash
curl -X POST https://<host>/api/auth/forgot-password \
     -H 'Content-Type: application/json' \
     -d '{"email": "<user>@<domain>"}'
# -> HTTP/1.1 404 Not Found
# -> {"message": "Not Found", "status": "error"}

curl -X POST https://<host>/api/auth/reset-password \
     -H 'Content-Type: application/json' \
     -d '{}'
# -> HTTP/1.1 400 Bad Request
# -> {"message": "token is required", "status": "error"}
```

---

## Related But Distinct

A second issue could be opened for the email-transport gap specifically (Manifest self-hosted does not support SMTP, and there is no documented `MANIFEST_EMAIL_*` environment variable family). That is a separate fix from this one and worth tracking independently for triage. Cross-link the two issues when both exist.

---

## Priority

High for self-hosted users. The current workaround (manual DB hash writes) will not survive a security audit, a staffing change, or a credential-already-leaked incident. The minimal implementation (forgot-password plugin plus a token-only mode with email logged rather than sent) would already be a substantial improvement.

guillaumegay13:
## What

On the password reset request page, detect when the server has **no email provider configured** and show a clear notice pointing to the authenticated Change Password flow, instead of the misleading "check your email" success that fires even when nothing was sent.

## Why (context: #2516)

Issue #2516 reports the reset flow as "broken" on self-hosted. Most of that report is a wrong endpoint path (`/api/auth/forgot-password` doesn't exist; the real routes are `/api/auth/request-password-reset` and `/api/auth/forget-password`), and email reset already works once `EMAIL_PROVIDER` + `EMAIL_API_KEY` are set.

The one genuine gap: when **no** email provider is configured, `sendEmail()` silently no-ops (logs a warning, returns `false`), so `requestPasswordReset` succeeds with a generic 200 and the reset page tells the user to "check your email" — but no email was ever sent. Better Auth calls `sendResetPassword` through `runInBackgroundOrAwait`, which swallows any throw, so a backend-side error can't surface here. The clean fix is a frontend notice driven by server config.

This pairs with #2514 (authenticated Change Password on the Account page), which is the recovery path this notice points users to.

## Changes

- **Backend**
  - `send-email.ts`: new `isEmailConfigured()` — the single source of truth for "will a reset email actually send" (`resolveConfig() !== null`).
  - `setup.service.ts`: `isEmailConfigured()` delegates to it.
  - `setup.controller.ts`: public `/api/v1/setup/status` now returns `emailConfigured`.
- **Frontend**
  - `setup-status.ts`: `checkEmailConfigured()` helper; **defaults to `true`** so a transient status fetch never hides the form — only an explicit `false` flips to the notice.
  - `ResetPassword.tsx`: request form reads the flag on mount and renders the notice (linking to `/account`) when email isn't configured.
  - `auth.css`: neutral `.auth-form__notice` styling.

## Tests

- Backend: `isEmailConfigured` unit tests (service + send-email), controller `emailConfigured` assertions, e2e contract check. 
- Frontend: `checkEmailConfigured` service tests + reset-page notice/subtitle/account-link tests.
- Full frontend suite (203 files) green; changed files at 100% line coverage; typecheck + lint clean.

Closes #2516

---
## Summary by cubic
Warn users on the Reset Password page when no email provider is configured, and point them to the Account page to change their password. Prevents the misleading “check your email” message on self-hosted installs without an email provider (addresses #2516).

- **New Features**
  - Backend: expose `emailConfigured` on `/api/v1/setup/status`, powered by a new `isEmailConfigured()` helper.
  - Frontend: `ResetPassword` calls `checkEmailConfigured()` on mount and shows a notice (with a link to `/account`) instead of the form when false.
  - UX: default to true when the backend omits the flag or on fetch errors, so transient issues don’t hide the form.

<sup>Written for commit a1320713795dd1ddbc2e1482127f25d90b1d688d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2519?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/reset-password-email-notice.md
+++ b/.changeset/reset-password-email-notice.md
@@ -1,0 +1,5 @@
+---
+"manifest": patch
+---
+
+The password reset page now detects when no email provider is configured and shows a clear notice (pointing to the authenticated Change Password flow) instead of silently pretending a reset link was sent. Self-hosted installs without an `EMAIL_PROVIDER` no longer dead-end here.
--- a/packages/backend/src/notifications/services/email-providers/send-email.spec.ts
+++ b/packages/backend/src/notifications/services/email-providers/send-email.spec.ts
@@ -2,8 +2,45 @@ jest.mock('./resolve-provider', () => ({
   createProvider: jest.fn(),
 }));
 
-import { sendEmail } from './send-email';
+import { sendEmail, isEmailConfigured } from './send-email';
 import { createProvider } from './resolve-provider';
+
+describe('isEmailConfigured', () => {
+  const originalEnv = process.env;
+
+  beforeEach(() => {
+    process.env = { ...originalEnv };
+    delete process.env['EMAIL_PROVIDER'];
+    delete process.env['EMAIL_API_KEY'];
+    delete process.env['EMAIL_DOMAIN'];
+    delete process.env['MAILGUN_API_KEY'];
+    delete process.env['MAILGUN_DOMAIN'];
+  });
+
+  afterEach(() => {
+    process.env = originalEnv;
+  });
+
+  it('returns false when no provider is configured', () => {
+    expect(isEmailConfigured()).toBe(false);
+  });
+
+  it('returns true for the unified EMAIL_* scheme', () => {
+    process.env['EMAIL_PROVIDER'] = 'resend';
+    process.env['EMAIL_API_KEY'] = 're_key';
+    expect(isEmailConfigured()).toBe(true);
+  });
+
+  it('returns false for mailgun without a domain', () => {
+    process.env['EMAIL_PROVIDER'] = 'mailgun';
+    process.env['EMAIL_API_KEY'] = 'key';
+    expect(isEmailConfigured()).toBe(false);
+  });
+
+  it('honors an explicit env override argument', () => {
+    expect(isEmailConfigured({ emailProvider: 'resend', emailApiKey: 're_key' })).toBe(true);
+  });
+});
 
 describe('sendEmail', () => {
   const originalEnv = process.env;
--- a/packages/backend/src/notifications/services/email-providers/send-email.ts
+++ b/packages/backend/src/notifications/services/email-providers/send-email.ts
@@ -63,6 +63,17 @@ function resolveConfig(env?: SendEmailEnvConfig): EmailProviderConfig | null {
   return null;
 }
 
+/**
+ * True when a transactional email provider is fully configured, i.e. a call to
+ * `sendEmail` would actually dispatch rather than no-op. This is the single
+ * source of truth for "can we send email?" — the setup status endpoint reads it
+ * so the password-reset UI can warn instead of silently pretending a reset link
+ * was sent on installs with no provider wired in.
+ */
+export function isEmailConfigured(env?: SendEmailEnvConfig): boolean {
+  return resolveConfig(env) !== null;
+}
+
 export async function sendEmail(
   opts: SendEmailOptions,
   env?: SendEmailEnvConfig,
--- a/packages/backend/src/setup/setup.controller.spec.ts
+++ b/packages/backend/src/setup/setup.controller.spec.ts
@@ -14,6 +14,7 @@ describe('SetupController', () => {
   let mockIsSelfHosted: jest.Mock;
   let mockIsOllamaAvailable: jest.Mock;
   let mockGetLocalLlmHost: jest.Mock;
+  let mockIsEmailConfigured: jest.Mock;
 
   beforeEach(async () => {
     mockNeedsSetup = jest.fn();
@@ -22,6 +23,7 @@ describe('SetupController', () => {
     mockIsSelfHosted = jest.fn().mockReturnValue(false);
     mockIsOllamaAvailable = jest.fn().mockResolvedValue(false);
     mockGetLocalLlmHost = jest.fn().mockReturnValue('localhost');
+    mockIsEmailConfigured = jest.fn().mockReturnValue(true);
 
     const module: TestingModule = await Test.createTestingModule({
       controllers: [SetupController],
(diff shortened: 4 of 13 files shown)


guillaumegay13:
Good catch, and it was valid against the branch's original base. This branch was cut from a stale `main` (before #2514), so `/account` had no Change Password flow at review time.

Fixed by rebasing onto current `main`, which now includes #2514 — the Account page has the credential-account Change Password form, so the notice link points at a real, implemented recovery path. No copy change needed.

Issue #2543, opened 2026-07-21
Knob merge sends thinking.budget_tokens alongside thinking.type: 'adaptive' → Anthropic 400; autofix reports the pre-translation body so Phoenix can't heal it


guillaumegay13:
## Symptom

Live traffic to `claude-opus-4-8` (inbound `/v1/responses` dialect) fails with:

```json
{
  "error": {
    "type": "invalid_request_error",
    "message": "thinking.adaptive.budget_tokens: Extra inputs are not permitted"
  },
  "statusCode": 400
}
```

The wire body Anthropic received carried `thinking: { type: "adaptive", budget_tokens: N }` — the legacy budget knob mixed into the new adaptive shape. Anthropic's `adaptive` union member doesn't permit `budget_tokens`.

## Root cause 1 — stale knob leaks past the applicability filter

`budget_tokens` is never written by Manifest source (zero hits outside `node_modules`); it comes from **stored route knobs** merged into the wire body by `applyRequestParamDefaults` (`packages/shared/src/request-params.ts`).

The modelparams catalog gates `thinking.budget_tokens` on `applicability.only: { "thinking.type": "enabled" }`, and `anthropic/claude-opus-4-8`'s spec has no budget knob at all (`max_tokens`, `thinking.type: disabled|adaptive`, `thinking.display`, `output_config.effort`). But older specs (e.g. `anthropic/claude-opus-4-subscription`) have **both** `adaptive` in the `thinking.type` enum **and** `thinking.budget_tokens` — so a budget saved under an older route survives alongside `type: "adaptive"` and `omitProviderInapplicableParams` / `providerParamIsApplicable` lets it through to the wire.

Fix candidates:
- enforce the `only: { thinking.type: "enabled" }` applicability at merge time so `budget_tokens` is dropped whenever the effective type isn't `enabled`;
- scrub stored knobs that are no longer in (or no longer applicable under) the route's current spec.

## Root cause 2 — autofix reports the wrong body to Phoenix

`AutofixService.maybeHeal` is called with `requestBody: body` (`packages/backend/src/routing/proxy/proxy.service.ts` — the **inbound client-dialect body**, before `toChatCompletionsRequest`, before the knob merge, before `toAnthropicRequest`). The contract says otherwise:

```ts
// packages/backend/src/routing/autofix/autofix.types.ts
/** The request body that was actually forwarded and failed. */
requestBody: Record<string, unknown>;
```

Consequences for the heal loop:
[...]

PR #2545, opened 2026-07-21
fix(routing): scrub stale catalog params stranded by the model-param merge (#2543)

guillaumegay13:
Part of #2543 — fixes **root cause 1** (the wire-shape defect). The autofix reporting gap (root cause 2) is left for a follow-up; see notes below.

## What production is actually doing

The failing requests are **Claude Code (`claude-cli`) speaking the Anthropic Messages dialect** (`/v1/messages`), not stored-knob-only traffic:

1. Claude Code sends the legacy extended-thinking shape: `thinking: {type: "enabled", budget_tokens: N}`.
2. The route's saved knobs for `anthropic/subscription/claude-opus-4-8` are `{thinking: {type: "adaptive", display: "omitted"}}`.
3. `applyRequestParamDefaults` deep-merges the knobs **over** the caller's `thinking` object — flipping `type` to `adaptive` while the caller's `budget_tokens` survives underneath.
4. `omitProviderInapplicableParams` cannot scrub it: `claude-opus-4-8`'s MPS spec has **no** `thinking.budget_tokens` path at all (only older specs like `claude-opus-4` carry it), and the scrubber only judges paths the resolved spec defines.
5. Wire: `{"type":"adaptive","budget_tokens":N,"display":"omitted"}` → Anthropic 400 `thinking.adaptive.budget_tokens: Extra inputs are not permitted`.

Confirmed against prod data: no `agent_model_params` row combines `adaptive` + `budget_tokens` (so the stored knobs are clean — the stale field rides in on the request), and the failing `agent_messages` rows interleave with successful ones on the same route depending on whether the caller had extended thinking on. Reproduced byte-for-byte with the real catalog.

## The fix

When the merge **rewrites** a value under a nested root (a body value at a spec path changed), caller-sent siblings under that root are dropped iff the **catalog-wide** param path set knows the path but the resolved model's spec does not. That is exactly the "param that belonged to the shape the caller chose, which no spec on this route can vouch for next to the merged values" case.

Deliberately narrow:
- **Unknown vendor extensions are preserved** (pins the existing `vendor_note` test).
- **Untouched roots are preserved** — if the merge doesn't change anything under `thinking`, the caller's body passes through as before (including a caller's own invalid combos: pass-through fidelity is not Manifest's to rewrite).
- **Spec-defined paths are untouched** — the existing applicability machinery (`only`/`except` + `omitProviderInapplicableParams`) keeps owning those.
- The known-path set is catalog-wide (not per-provider) so custom providers proxying Claude models — which have the same stored `enabled`+`budget` rows in prod — are covered too.

## Changes

- `packages/shared/src/request-params.ts` — `applyRequestParamDefaults` gains an optional `knownParamPaths` set + `dropStaleCatalogSiblings`.
- `packages/backend/src/routing/routing-core/provider-param-spec.service.ts` — `knownParamPaths()`: catalog-wide path set, computed once.
- `packages/backend/src/routing/proxy/proxy-fallback.service.ts` — the per-attempt merge passes the set.
- Tests: 8 new shared cases (incl. the exact production repro against an opus-4-8-shaped spec) + fallback-service regression test + service test.

Backend: 413 suites / 7,560 tests green. Shared: 360 tests green, 100% line coverage kept (branch coverage on `request-params.ts` improved 92.85% → 96.42%).

## Root cause 2 (not in this PR)

Autofix still reports the pre-translation inbound body to Phoenix. The clean fix is to attach the built wire body (`BuiltProviderRequest.requestBody`) to `ForwardResult`, report that, and resend a healed wire body verbatim (skipping re-merge/re-translation). It needs a contract decision first: Phoenix's `api` enum (`chat_completions|responses|messages`) can describe Anthropic/OpenAI-shaped wire bodies but has no member for Google-format upstreams. Happy to do that as a follow-up once the Phoenix side is settled.

---
## Summary by cubic
Fixes Anthropic 400s by scrubbing stale caller params (e.g. `thinking.budget_tokens`) when the model-param merge rewrites a nested root (e.g. flips `thinking.type` to `adaptive`) — addresses #2543 root cause 1. Uses a catalog-wide param path set to drop paths the catalog knows but the resolved model’s spec no longer defines, preventing failures on `claude-opus-4-8` while preserving vendor extensions and unchanged roots.

- **Bug Fixes**
  - Added `knownParamPaths()` and passed it to `applyRequestParamDefaults` to drop stale siblings under rewritten roots.
  - Keeps unknown vendor extensions and untouched roots; spec-defined paths remain governed by applicability; works for custom providers too.
  - Added tests reproducing the prod failure and regression coverage across routing and shared utilities.

<sup>Written for commit 27b142c1ce2b8e83667f95ae8f4c2b0bf7df6869. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2545?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/stale-thinking-budget-scrub.md
+++ b/.changeset/stale-thinking-budget-scrub.md
@@ -1,0 +1,5 @@
+---
+'manifest': patch
+---
+
+Fix Anthropic 400 "thinking.adaptive.budget_tokens: Extra inputs are not permitted": when configured model params rewrite a nested request root (e.g. flip `thinking.type` to `adaptive`), caller-sent siblings that the MPS catalog knows but the resolved model's spec no longer defines (like a legacy `thinking.budget_tokens`) are now scrubbed from the outbound body instead of riding into a wire shape the provider rejects.
--- a/packages/backend/src/routing/proxy/__tests__/proxy-fallback.service.spec.ts
+++ b/packages/backend/src/routing/proxy/__tests__/proxy-fallback.service.spec.ts
@@ -34,6 +34,60 @@ const specCatalog: ProviderParamSpecCatalog = [
       },
     ],
   },
+  // Older Anthropic generation: knows both `enabled` and `thinking.budget_tokens`.
+  {
+    provider: 'anthropic',
+    authType: 'subscription',
+    model: 'claude-opus-4',
+    params: [
+      {
+        path: 'thinking.type',
+        type: 'enum',
+        label: 'Thinking mode',
+        description: 'Controls Anthropic thinking mode.',
+        default: 'disabled',
+        values: ['disabled', 'adaptive', 'enabled'],
+        group: 'reasoning',
+      },
+      {
+        path: 'thinking.budget_tokens',
+        type: 'integer',
+        label: 'Budget tokens',
+        description: 'Extended thinking token budget.',
+        default: 4096,
+        range: { min: 1024 },
+        group: 'reasoning',
+        applicability: { only: { 'thinking.type': 'enabled' } },
+      },
+    ],
+  },
+  // Current generation: adaptive-only, no budget param at all.
+  {
+    provider: 'anthropic',
+    authType: 'subscription',
+    model: 'claude-opus-4-8',
+    params: [
+      {
+        path: 'thinking.type',
+        type: 'enum',
+        label: 'Thinking mode',
+        description: 'Controls Anthropic thinking mode.',
+        default: 'disabled',
+        values: ['disabled', 'adaptive'],
+        group: 'reasoning',
+      },
+      {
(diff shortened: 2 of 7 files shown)


guillaumegay13:
Superseded by #2552, which applies the same fix locally at the Anthropic wire boundary without changing shared parameter-merge behavior.

PR #2565, opened 2026-07-24
feat: record messages per provider attempt

guillaumegay13:
### ✨ What changed
- Store exact provider payloads outside PostgreSQL per Provider Attempt.
- Show Messages inside the selected Attempt.
- Use S3 when fully configured, otherwise use persistent filesystem storage.
- Add a safe Compose upgrade path for the recordings volume.
- Seed a realistic OpenAI attempt recording for local development.

### 💭 Why
Request-level payloads blur Auto-fix retries and fallbacks. Each provider call needs its own wire exchange.

### 👤 For users
Enabled agents show new messages on each Attempt. Existing requests are unchanged.

### 🔧 For operators
The migration adds nullable `agent_messages.recording_key`; payloads stay in S3 or the filesystem.
Compose uses a named volume. Existing installs run `bash install.sh --upgrade` once.
Complete S3 variables select object storage automatically.
Objects use `request-recordings/v2/tenants/<tenant>/requests/<request>/attempts/<attempt>.json.gz`.

### 📝 Notes
Validated on a sanitized clone of 7,536,190 production Provider Attempts with zero row changes.


SebConejo:
@guillaumegay13 
1. The messages are related to the request but they are displayed in a message
2. The content is hard to undestadn, in the User card, we ca nread the content from the assistant. The assistant content should be displayed in the assistant card.

PR #2607, opened 2026-07-28
feat: add self-hosted Autofix instance identity

guillaumegay13:
### ✨ What changed
- Self-hosted installs heal against hosted Autofix with no Phoenix configuration. With no `AUTOFIX_HEALING_API_KEY` set, calls carry `X-Manifest-Instance` (the anonymous `install_metadata.install_id`, minted lazily) plus `X-Manifest-Version` and `X-Manifest-Harness`. Cloud's static-key path is unchanged and omits the instance headers.
- The instance id is an **identifier, not a credential**: not secret, not encrypted, and there is no registration handshake and no 401 refresh — Phoenix creates the instance row the first time it sees an id.
- Self-hosted consent-once before Autofix turns on (`install_metadata.autofix_consented_at`), surfaced in the sidebar CTA, per-agent settings, and first agent creation.
- Adds `GET /api/v1/autofix/status` → `{ any_enabled, enabled_agents, needs_enable_all, consented }` and `POST /api/v1/autofix/enable-all`.
- The healer URL is now the `AUTOFIX_URL` constant rather than the `AUTOFIX_HEALING_URL` env var. Production — cloud and self-hosted alike — heals against it; dev/test use the in-process mock.

### 💭 Why
Turning on Autofix for a self-hosted agent should work without manual Phoenix configuration.

### 🔧 Notes
- Adds migration `1801900000000-AddAutofixConsentToInstallMetadata`. [Phoenix support](https://github.com/mnfst/phoenix/pull/161) is already deployed.
- **The opt-out is now `AUTOFIX_GLOBAL_ENABLED=false`.** `AUTOFIX_HEALING_URL` and its `off` sentinel are removed. That flag now also skips the boot health probe, so it means no calls to the Autofix service at all — previously `off` carried that guarantee while the flag left the probe running.
- **`enable-all` enables every live, non-playground agent, including any previously switched off** — it is not limited to unconfigured ones. `needs_enable_all` only gates whether the CTA is shown (no agent enabled *and* at least one still `NULL`). The cubic summary below states that explicit per-agent choices are preserved; that is inaccurate.
- Cloud production set `AUTOFIX_HEALING_URL` to this same host in Railway. The constant deliberately covers cloud as well, so removing the variable does not send cloud to the inert client — pinned by a test.
- Removes `NoopHealingClient`: it existed only for "production with no healer URL configured", a state that can no longer occur.

---
## Summary by cubic
Adds self‑hosted Auto‑fix instance identity with inline, once‑per‑install consent and a per‑agent enable modal that surfaces disabled agents and records consent on first enable. Production heals against hosted `AUTOFIX_URL`; dev/test use the in‑process mock; opt‑out via `AUTOFIX_GLOBAL_ENABLED=false`.

- **New Features**
  - Per‑agent enable modal: Sidebar card shows while any live agent runs without Auto‑fix; lists uncovered agents with independent toggles, session‑dismissible; first enable sets install consent. Status adds `disabled_agents` and `consented`, refetches on SSE pings plus a 15s poll. Copy and layout polish: card leads with reliability, header alignment fixed, modal title made affirmative, list label shortened to “Agents”.
  - Inline consent: No confirmation dialog; the legal line under toggles is the consent act. Per‑agent PATCH/GET include `consented`. Fleet `POST /api/v1/autofix/enable-all` stays for API backfill.
  - “Logs” vocabulary and create settings: Add‑agent and Welcome include “Auto‑fix” and “Enable logs” switches (both default ON). Create sends `autofix_enabled`/`record_messages`.
  - Hosted healer constant: removed `AUTOFIX_HEALING_URL`. Prod always targets https://autofix.manifest.build; dev/test resolve to the mock. Calls send `X-Manifest-Instance` (anonymous install id) plus `X-Manifest-Version` and `X-Manifest-Harness` when no `AUTOFIX_HEALING_API_KEY` is set.

- **Bug Fixes**
  - Cache and status: Bust `/autofix/status` on both enable and disable; keep last known status on failed polls; sidebar soft‑handles errors.
  - Consent races: Guard Add‑agent create with an attempt token and a `checkingConsent` lock; prevent stale redirects when the modal reopens; focus‑trap/layout polish.
  - Gating: `needs_enable_all` requires legacy/unconfigured agents and a never‑consented install; backend skips soft‑deleted agents.

<sup>Written for commit d220958d27ce9b84a43939fd706783091519aaca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2607?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


guillaumegay13:
Heads-up on two things in this PR that are worth a reviewer's attention.

**1. `enable-all` is broader than the cubic summary says.** The auto-generated summary states it "backfills only legacy/unconfigured agents; explicit per-agent choices are preserved". It doesn't — `AutofixStatsService.enableAll()` runs:

```ts
await this.agentRepo.update(
  { tenant_id: tenantId, is_playground: false, deleted_at: IsNull() },
  { autofix_enabled: true },
);
```

That covers **every** live non-playground agent, including ones a user had deliberately switched off. `needs_enable_all` only decides whether the CTA is *shown* (no agent enabled **and** at least one still `NULL`), not what the update touches. I've corrected the PR description; flagging it here because cubic regenerates its own block on each new commit. If preserving explicit opt-outs was the intent, the fix is a `autofix_enabled IS NULL` predicate on the update.

**2. The healer URL is now a constant, which changes cloud too.** `AUTOFIX_HEALING_URL` is gone; production resolves the `AUTOFIX_URL` constant instead. Cloud production had that variable set in Railway to this same host, so resolving the constant for self-hosted only would have dropped cloud to the inert client and silently disabled healing for hosted tenants — the constant covers both modes and there's a test pinning it.

Two consequences worth confirming:
- The documented opt-out changed. `AUTOFIX_HEALING_URL=off` no longer exists; `AUTOFIX_GLOBAL_ENABLED=false` is the only switch, so it now also skips the boot health probe (it previously left the probe running, which was fine while `off` guaranteed zero contact). DOCKER_README updated to match.
- `NoopHealingClient` is deleted — it only served "production with no healer URL configured", which can't happen now.

Verified against the real deployed healer: a repairable 400 drove `maybeHeal` → `POST https://autofix.manifest.build/api/heal` → recorded as `requests.autofix_status='no_patch'` with the Phoenix `issueId` on the attempt row, no patched retry, client got the original provider error.

Issue #2571, opened 2026-07-25
Constant M500 Internal Error on Self Hosted


omaquu:
Getting constantly M500 Error, also it seems it doesnt always try to go for fallback.

ERROR [ProxyController] Proxy error: This operation was aborted

guillaumegay13:
Hi @omaquu, I improved error handling. Will be available on the next docker version.

rondlite:
> Getting constantly M500 Error, also it seems it doesnt always try to go for fallback.
> 
> ERROR [ProxyController] Proxy error: This operation was aborted

In case you on k8s and using bitnami helm chart for your postgres.. check that resource size is not set to nano.
Had 500s due to postgres crashing for that reason.

guillaumegay13:
@omaquu any news?

PR #2735, opened 2026-08-19
Add pagination to usage-based providers list

### LINKED ISSUE #2268: Provider list: Make sure users can see the list of supported providers to add them
For current users, the list of supported providers appears at the bottom of the page after a huge list of connected providers.

We could improve the UI to make sure it's easy to understand where to add new providers. 

Suggestions:
- Add pagination to the connected providers (10 or 15 per page)
- Add filters to to the connected providers
- Allow to sort connected providers
- Make the default order by usage so that I see the most used connected providers by default
- Or add tabs

HarveyPunsalan:
Fixes #2268

Adds pagination to the usage-based providers list (both grid and list views), starting with the pagination suggestion from the issue. Uses the existing createClientPagination utility and Pagination component already in the codebase. Page size set to 20. Tested locally, all existing tests still pass.

---
## Summary by cubic
Adds client-side pagination (20 items/page) to the usage-based providers list in both table and grid views. Previously all providers rendered at once; now only the current page renders with Next/Previous controls, improving render performance with no server or API changes.

- Uses `createClientPagination` over `providerListForKind(kind)`; both views render `pager.pageItems()` and react when the kind changes.
- Adds `Pagination` below both table and grid; wires `currentPage`, `totalItems`, `pageSize`, `hasNextPage`, `onPrevious`, and `onNext`.
- Fixes grid layout by placing pagination outside the grid so controls appear below the cards.

<sup>Written for commit 522a5ad8668fc18aff18081104489c44d2d1f816. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2735?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


SebConejo:
@HarveyPunsalan Thanks a lot for picking this one up, and for wiring it through both views.

I pulled the branch and ran it locally, and I think there's a mix-up on **which** list got paginated. I'd rather flag it now than after merge.

## The list that needs pagination is the connections one

The problem in #2268 is that people can't find where to add a new provider: the "Supported providers" catalog sits at the very bottom of the page, behind a connections table that gets long. You had it right in your comment on the issue, you said you'd start with the connected providers list.

The code paginates the other one though. Both `<For>` loops that changed (lines 842 and 899) render `providerListForKind(props.kind)`, which is the supported catalog under `{copy().supportedHeading}`. `connectedRows()` at line 613 is untouched.

So the connections table is exactly as long as before, the catalog is still just as far down the page, and now the catalog itself is truncated. Concretely, on current `main` the BYOK catalog has 24 entries, so page 1 shows 20 and **Pioneer, xAI, Xiaomi MiMo and Z.ai now need a click on Next to be seen**. On Subscriptions (18 providers) and Local (3) nothing changes at all, both sit under the page size. So the feature only ever activates on one page, and its only visible effect there is hiding four providers.

Moving the pager onto `connectedRows()` would address the actual complaint.

## Two other things

**Patch coverage is red.** `codecov/patch` fails at 88.89%, 3 lines missing: the `currentPage`, `hasNextPage` and `onPrevious` props on the list-view `<Pagination>` (886, 889-890). They're never evaluated because the only list-view test renders `<Subscriptions />`, which has 18 providers and stays below the threshold, so the `<Show>` inside `Pagination` never opens. A test that renders `<Byok />`, switches to list view and clicks Next would cover them. We keep patch coverage at 100% on this repo.

**Grid view puts the pagination outside any panel.** Its CSS (`styles/data.css:662`) gives `.pagination` a `border-top` and padding because it's designed as a table footer, and both existing call sites keep it inside a `.panel` right under a `<table>` (`ModelPrices.tsx:356`, `MessageLog.tsx:999`). As it stands it renders as a floating rule under the cards. Wrapping the grid and the pagination together in a panel would do it.

## A recommendation

Thinking about it more, I'm not convinced pagination is the right shape here at all. Paginating one section in the middle of a page feels odd to me, and it hides rows people may be looking for.

What I'd rather have on the connections section: cap it at around 10 rows of height, make it scroll inside its own card, and put a "Show all" button underneath that expands it to full height. Nothing is ever hidden, every connection stays reachable by scrolling inside the card, and the supported catalog moves up to a reachable spot straight away, which is the whole point of the issue.

If that sounds good to you, go for it on this branch. Also worth merging latest `main` in, you're 8 commits behind and one of them adds a new provider.

SebConejo:
@HarveyPunsalan 

Another PR solved this issue: #2571

**Why I didn't chose to create pagination.**

This page list several things: Key numbers, connections, and available providers. 

Adding a pagination to one section was a bit tricky. It could work but as you seen, it was not obvious to make sure it's easier to use for each user. What will change after clicking on the pagination? The connections or the prfviders? Moreover, maybe a user could see the whole connection list. 

**The solution I implemented**
We display now a maximum of 6 - 7 connexions, with an expend button. The button allows you to see all your connections.

So immediately you have an overview of everything. You can expand the connexions if you wnat or directly go thorugh the available providers.

PR #2378, opened 2026-07-01
Show subscription usage limits

ziahamza:
## Summary
- add a tenant-scoped subscription usage endpoint for OpenAI/Codex, Anthropic/Claude, and Gemini OAuth connections
- normalize live provider quota windows into per-connection usage details, combining multiple installed connections under each provider summary
- show the limit details in the Subscriptions page under My subscription connections

## Validation
- npm --workspace manifest-backend test -- subscription-usage.service.spec.ts --runInBand
- npm --workspace manifest-backend run build
- npm --workspace manifest-frontend run build
- npx eslint packages/backend/src/routing/subscription-usage.service.ts packages/backend/src/routing/subscription-usage.controller.ts packages/backend/src/routing/subscription-usage.service.spec.ts packages/backend/src/routing/routing.module.ts packages/frontend/src/services/api/providers.ts packages/frontend/src/pages/providers/ProviderConnectionsPage.tsx
- E2E QA against a separate Docker Compose QA container on 127.0.0.1:3099 with a real account: subscriptions page rendered live Anthropic, Gemini, and OpenAI limits and /api/v1/providers/subscription-usage returned all three providers with status ok

---
## Summary by cubic
Show live subscription limits for `openai`, `anthropic`, `gemini`, and `xai` OAuth connections with compact meters on the Subscriptions page and a detailed limits panel on each connection’s detail view. Subscriptions combine multiple connections into one row per provider; gauges are color‑coded by projected pace and responsive on mobile.

- New Features
  - Backend (`packages/backend`): Adds tenant‑scoped `GET /api/v1/providers/subscription-usage` with `xai`/Grok support; validates gRPC‑web frames/trailers, parses raw/framed protobuf, maps gRPC status to HTTP, and normalizes used%, remaining%, reset time, and window (5h/daily/weekly/monthly) into a shared schema; flags inactive/missing/expired creds as `unavailable`; provider status is `partial` when mixed and `unavailable` when all connections are unavailable; 12s timeout; returns `{ providers: [...] }`; tests cover transport errors/timeouts, Grok parsing (incl. trailer‑only), malformed protobuf, tenant scoping, exact upstream hosts, controller response wrapping, and summary `unavailable`.
[...]

PR #2739, opened 2026-08-19
feat: per-route "skip on quota exhaustion" toggle for subscription providers

tunwinlat:
## Summary

Subscription providers with usage windows (Anthropic, Kimi, OpenAI, MiniMax, xAI) currently keep getting attempted on every request even when their quota is exhausted — each attempt fails with a quota/rate-limit error before the fallback chain engages, adding avoidable latency to every routed request.

This PR adds an opt-in **per-route "Skip on quota exhaustion"** flag: a new backend service polls each subscription connection's usage endpoint on an interval, and the router skips flagged routes while their connection is known to be exhausted.

### What changed

- **`ModelRoute` gains `skipWhenQuotaExhausted?: boolean`** (shared type; stored in the existing jsonb route columns — **no migration**). A `quotaCheck` capability on the shared subscription configs marks which providers expose a usable usage endpoint.
- **New `SubscriptionQuotaService`** (`routing/subscription-quota.service.ts`): polls every 60s (`SUBSCRIPTION_QUOTA_POLL_INTERVAL_MS`, clamped to ≥30s), in-memory state only, **fail-open everywhere** — fetch errors, unwrap failures, or missing data never skip a route, and a failed refresh keeps the previous state. Per provider:
  - Anthropic: `GET api.anthropic.com/api/oauth/usage` (5h/weekly/per-model windows, `utilization >= 100`)
  - Kimi: `GET api.kimi.com/coding/v1/usages` (5h/weekly/monthly, `used >= limit`)
  - OpenAI: `GET chatgpt.com/backend-api/wham/usage` (`used_percent >= 100`)
  - MiniMax: `GET <resource-origin>/v1/token_plan/remains` (remaining-percent `<= 0`)
  - xAI: `POST grok.com/grok_api_v2.GrokBuildBilling/GetGrokCreditsConfig` (gRPC-web + dependency-free protobuf scan)
  - OAuth tokens reuse the same unwrap/refresh path the proxy uses at request time.
- **`ResolveService`** skips flagged routes backed by an exhausted connection in both promotion paths (tier chain and header tiers), for **primaries and fallbacks alike**. A **never-empty safeguard** keeps the original chain when every candidate is exhausted, so a harness is never left unroutable.
- **Routing UI**: a compact gauge-icon toggle (matching the existing Parameters affordance) on quota-capable subscription routes — tier overrides, fallback rows, specificity and header-tier cards. Default off.

### For users

New "Skip on quota exhaustion" toggle on subscription routes in a harness's routing. When enabled, an exhausted connection's models are bypassed until the quota window resets, so requests reach a working fallback immediately.

### For operators

- No migration, no new dependencies, in-memory state only (re-polls from scratch at boot).
- Optional env knob `SUBSCRIPTION_QUOTA_POLL_INTERVAL_MS` (default 60000, min 30000).

### Notes

- **Kimi's `/coding/v1/usages` is undocumented** (reverse-engineered, used by several community tools); the parser is deliberately defensive (string numbers, window fallbacks).
- **The xAI parser relies on protobuf field heuristics** over gRPC-web (grok.com has no public usage API). It fails open if the schema changes. A regression fixture captured from a real live response is included in the spec.
- **Relation to #2378**: that PR adds a display-only usage endpoint; this one feeds routing decisions. Intentionally a standalone service (`subscription-quota.*`, not `subscription-usage.*`) to avoid conflicting with the unmerged PR — the fetchers could be unified later.
- `supportsQuotaCheck` (shared configs, drives UI visibility) and the backend service's provider list are two sources of truth to keep in sync when adding providers.

### Validation

- `npm test --workspace=packages/shared` — 429/429
- `npm test --workspace=packages/backend` — full suite 458/458 suites, 8413 tests
- Frontend: all touched/neighboring suites green (Routing, FallbackList, tier cards, header tiers)
- `npm run build` — clean; eslint clean on all changed files
- Live-verified all five providers' usage endpoints against real subscription connections on a self-hosted install, including the skip behavior end to end

---
## Summary by cubic
Skips exhausted subscription routes during routing so requests hit working fallbacks faster. Previously every candidate was attempted even when its subscription was exhausted; now quota-capable routes can opt in to be skipped, with a safeguard that keeps chains non-empty.

- Adds skipWhenQuotaExhausted?: boolean to ModelRoute (stored in existing jsonb; no migration). PUT omitting the field preserves the stored flag; explicit false clears it; specificity swaps and fallback promotion retain it. DTOs whitelist the flag; tier, specificity, and header-tier controllers forward it.
- Introduces a quota poller service (`SubscriptionQuotaService` via `SubscriptionQuotaModule`) that polls Anthropic, Moonshot (Kimi), OpenAI, MiniMax, and xAI every 60s (`SUBSCRIPTION_QUOTA_POLL_INTERVAL_MS`, min 30s; digits-only). In-memory, fail-open; failed refresh keeps prior state; prunes deleted/disabled connections; reuses existing OAuth refresh.
- Router consults quota state for primaries and fallbacks across default tiers, specificity overrides, and header tiers; if all candidates are exhausted it keeps the original chain.
- UI adds a compact gauge-icon toggle on quota-capable subscription routes (primaries and fallbacks, including header tiers). Default off. Guards double-click races and in-flight saves.
- `manifest-shared` exposes `supportsQuotaCheck` and marks provider coverage. Docs and tests updated; no new runtime dependencies.

<sup>Written for commit ec6f4efa62c76120124e113fc468a1e9dd8b7676. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2739?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/quota-aware-routing-skip.md
+++ b/.changeset/quota-aware-routing-skip.md
@@ -1,0 +1,5 @@
+---
+"manifest": minor
+---
+
+Add per-route "skip on quota exhaustion" toggle for Anthropic, Kimi, OpenAI, MiniMax, and xAI subscription providers
--- a/docs/providers/subscription-based-providers.md
+++ b/docs/providers/subscription-based-providers.md
@@ -33,6 +33,10 @@ If you already pay for one of the plans below, Manifest can route through that s
 
 Providers that support both API keys and subscriptions show both tabs. Subscription-only providers show only the subscription flow. In routing pickers, subscription rows appear only after a usable subscription credential is saved (`has_api_key=true`).
 
+## Skip on quota exhaustion
+
+For providers with a usage endpoint — Anthropic (Claude), Moonshot (Kimi Coding Plan), OpenAI (ChatGPT), MiniMax, and xAI (Grok) — each route in a tier's primary slot or fallback chain offers a "Skip on quota exhaustion" toggle (the gauge icon). When enabled, Manifest polls the provider's quota state and skips that route while the subscription is exhausted, falling through to the next route instead of burning a failed attempt. Off by default; the poll interval defaults to 60s (`SUBSCRIPTION_QUOTA_POLL_INTERVAL_MS`, minimum 30s).
+
 ## How auth works per provider
 
 ### OpenAI
--- a/packages/backend/src/routing/__tests__/specificity.controller.spec.ts
+++ b/packages/backend/src/routing/__tests__/specificity.controller.spec.ts
@@ -72,8 +72,35 @@ describe('SpecificityController', () => {
         'openai',
         'api_key',
         undefined,
+        undefined,
       );
       expect(result).toBe(override);
+    });
+
+    it('should forward skipWhenQuotaExhausted from the structured route', async () => {
+      const body = {
+        model: 'claude',
+        route: {
+          provider: 'anthropic',
+          authType: 'subscription' as const,
+          model: 'claude',
+          skipWhenQuotaExhausted: true,
+        },
+      };
+      mockSpecificityService.setOverride.mockResolvedValue({ id: 'sa-1' });
+
+      await controller.setOverride(ctx, 'test-agent', 'coding', body);
+
+      expect(mockSpecificityService.setOverride).toHaveBeenCalledWith(
+        'agent-1',
+        'tenant-1',
+        'coding',
+        'claude',
+        'anthropic',
+        'subscription',
+        undefined,
+        true,
+      );
     });
 
     it('should throw BadRequestException for invalid category', async () => {
--- a/packages/backend/src/routing/dto/routing.dto.ts
+++ b/packages/backend/src/routing/dto/routing.dto.ts
@@ -39,6 +39,13 @@ export class ModelRouteDto {
   @MaxLength(MAX_PROVIDER_KEY_LABEL_LENGTH)
   @Transform(({ value }) => (typeof value === 'string' ? value.trim() : value))
   keyLabel?: string;
+
+  // Opt-in per-route flag: the router skips this route while the backing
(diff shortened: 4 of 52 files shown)


guillaumegay13:
hi @tunwinlat, thanks for your contribution!

before moving forward, Manifest already has a cooldown system where it'll skip the route : 15s default for rate limits and 5 min max based on the Retry-After value in the error response.

it's not enough for you?

tunwinlat:
> hi @tunwinlat, thanks for your contribution!
> 
> before moving forward, Manifest already has a cooldown system where it'll skip the route : 15s default for rate limits and 5 min max based on the Retry-After value in the error response.
> 
> it's not enough for you?

Hi @guillaumegay13, thanks for the attention. Sorry, I missed some contexts in the summary.

I am trying to fix 2 things that I find a bit bothersome with retrying when quota is full on top of just failed attempts.

1. when the provider has extra usage enabled with credits, I am not able to restrict some harnesses that are token hungry to stop using the provider. this allows me to skip the model with exhausted quota for the harnesses I want.
2. fallback caused by quota exhaustion contributes to the error rate. It's misleading because while it's an error in a way, that's an expected behaviour when quota is exhausted. But a global behaviour change is challenging because some providers don't support that. So I wanted to shoot all 3 things in one go.

Please let me know if you have any questions or suggestions.

PR #2747, opened 2026-08-20
feat(admin): scoped AI-admin API surface (mnfst_admin_ai_*) for self-hosted management

sponsorrockit1:
## What

Adds a **scoped AI-admin API** under `/api/v1/admin` for programmatically
administering a self-hosted Manifest instance — agents, provider keys, and
observability — gated by a new key scope (`ai_admin`) on a dedicated
`mnfst_admin_ai_*` key prefix.

This is a thin, additive layer over the existing NestJS service layer
(`AgentLifecycleService`, `ProviderService`, `CustomProviderService`,
`TimeseriesQueriesService`, `ResolveAgentService`). **No new business logic,
no new infrastructure, no breaking changes.**

## Why

Self-hosted instances are currently managed almost entirely through the
dashboard (browser session / owner key). There is no stable, key-authenticated
*machine* surface for an AI agent or ops automation to onboard and operate
harnesses without a human in the browser. An AI-admin key is **narrowly scoped**
so it can administer agents/providers/observability but, by construction, cannot
hit owner-only account/session endpoints.

## Changes

- `api_keys.scope` column (`owner` default, `ai_admin`) + migration.
- `tenant_providers.key_hash` column + migration (enables provider-key
  **match-verify** without ever returning the raw secret).
- `ApiKeyGuard` stashes the resolved scope on the request; new `AdminAiGuard`
  authorizes `/api/v1/admin/*` to `ai_admin` only (owner keys → 403, missing → 401).
- `AdminController` — mint/list/revoke admin keys (raw key returned once).
- `AdminAgentController` — full agent CRUD + `rotate-key` + `duplicate`.
- `AdminProviderController` — list + attach/update provider key +
  **match-verify** (`{ match: boolean }`, salt-aware `verifyKey`, secret never returned).
- `AdminObservabilityController` — read-only usage + per-agent routing.
- Exported the needed services from `AnalyticsModule`/`RoutingCoreModule`/
  `CustomProviderModule` (proper Nest boundaries) so `AdminModule` imports them
  rather than re-providing.
- Registered the two new migrations in `data-source-definitions.ts` (mnfst
  registers migrations as an explicit array, not a glob).
- Changeset included.

## Security

- No handler returns a stored provider/agent secret; ingest keys are returned
  only at creation/rotation/duplicate (same as the dashboard).
- Match-verify uses salt-aware `verifyKey`, never a raw compare.
- Migrations are additive with defaults; existing rows untouched.

## Verification

- `nest build` green.
- Existing `custom-provider.controller.spec.ts` + `provider-key.service.spec.ts`
  = 22/22 pass (no regression).
- Built into a local image, deployed to a **live self-hosted instance**, and
  exercised end-to-end: agent create/get/rename/duplicate/rotate/delete,
  provider attach + match-verify (both directions), observability, and scope
  enforcement (200 with admin key, 401 without, 403 with owner key). Live proxy
  stayed healthy; real agents untouched.

## Discussion

Proposal + design questions for maintainers:
https://github.com/mnfst/manifest/discussions/2746

## Notes for review

- This is the agent-admin twin of the dashboard's existing controllers; the
  dashboard itself is unchanged.
- Per-project convention, I added a changeset rather than relying on release
  tooling to detect the change.

Happy to adjust scope/shape (prefix, scope modeling, match-verify placement) to
match project conventions before merge.

---
## Summary by cubic
Adds a scoped `/api/v1/admin` API for self-hosted agent, provider, and observability management gated by `mnfst_admin_ai_*` keys. Previously only the dashboard with owner keys could do this; now admin keys can do it programmatically, and owner keys are rejected on `/api/v1/admin` except `POST /api/v1/admin/keys` to bootstrap the first admin key.

- API surface: mint/list/revoke admin keys; agent CRUD (create/get/rename/rotate-key/duplicate/delete; GET key returns prefix only; raw keys only on create/rotate/duplicate); provider list/attach/update and key match-verify (no secret returned); read-only usage and routing.
- Data/auth: adds `api_keys.scope` (defaults to `owner`) and `tenant_providers.key_hash`; provider writes persist `key_hash`, and match-verify lazily backfills it for legacy encrypted rows; `ApiKeyGuard` stashes `authScope`, `AdminAiGuard` requires `ai_admin`, and `@AdminBootstrap` allows an owner key only for `POST /api/v1/admin/keys`.
- Notable behavior: owner keys get 403 on admin routes (bootstrap route allowed); duplicate agent names return 409; agent creation compensates if provider enablement fails; migrations include safer `down` lock timeouts; observability queries run in parallel.

**Rollout**
- Run DB migrations to add `api_keys.scope` and `tenant_providers.key_hash`.
- Bootstrap: use an owner key to `POST /api/v1/admin/keys` once to mint a `mnfst_admin_ai_*` key, then use that key for all `/api/v1/admin` calls.
- To enable match-verify for existing provider connections, re-save the provider key once (or rely on lazy backfill triggered by the first verify).

<sup>Written for commit d2e8bfc83015fbe51adcc76098b66972b806fc72. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2747?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/agent-admin-api.md
+++ b/.changeset/agent-admin-api.md
@@ -1,0 +1,12 @@
+---
+'manifest': minor
+---
+
+Add a scoped AI-admin API surface (`/api/v1/admin`) for agent-native administration of a self-hosted Manifest install.
+
+- New `scope` column on `api_keys` (defaults to `owner`, additive — existing keys unchanged).
+- New `mnfst_admin_ai_*` key type minted as `api_keys` rows with `scope = 'ai_admin'`.
+- `AdminAiGuard` restricts the admin surface to those keys; the global `ApiKeyGuard` now stashes the resolved scope on the request.
+- `AdminController` manages admin keys (create/list/revoke) and a health endpoint, reusing existing `ApiKey`/`Tenant` services.
+
+This is the foundation for agents to administer agents, providers, and routing via API (a wanted capability for the "AI Agents that don't break" mission). Subsequent milestones (M2–M4) layer agent CRUD, provider-key attach + match-verify, and routing/usage read over the existing services.
--- a/.changeset/agent-admin-api.md
+++ b/.changeset/agent-admin-api.md
@@ -4,9 +4,13 @@
 
 Add a scoped AI-admin API surface (`/api/v1/admin`) for agent-native administration of a self-hosted Manifest install.
 
-- New `scope` column on `api_keys` (defaults to `owner`, additive — existing keys unchanged).
+- New `scope` column on `api_keys` (defaults to `owner`, additive — existing owner keys unchanged) and a `key_hash` column on `tenant_providers` (one-way hash of the stored provider key, for match-verify without ever returning the raw secret).
 - New `mnfst_admin_ai_*` key type minted as `api_keys` rows with `scope = 'ai_admin'`.
 - `AdminAiGuard` restricts the admin surface to those keys; the global `ApiKeyGuard` now stashes the resolved scope on the request.
-- `AdminController` manages admin keys (create/list/revoke) and a health endpoint, reusing existing `ApiKey`/`Tenant` services.
+- Controllers under `/api/v1/admin`:
+  - `AdminController` — admin key create/list/revoke + health.
+  - `AdminAgentController` — full agent CRUD (create, get, rename, rotate-key, duplicate, delete) over the existing lifecycle services.
+  - `AdminProviderController` — attach/update a provider key (custom + standard) and **match-verify** a posted key against the stored one-way hash (no raw secret returned).
+  - `AdminObservabilityController` — read-only per-agent/per-provider usage and agent routing visibility.
 
-This is the foundation for agents to administer agents, providers, and routing via API (a wanted capability for the "AI Agents that don't break" mission). Subsequent milestones (M2–M4) layer agent CRUD, provider-key attach + match-verify, and routing/usage read over the existing services.
+This is the foundation for agents to administer agents, providers, and routing via API (a wanted capability for the "AI Agents that don't break" mission). Provider-key writes now also persist `key_hash` so match-verify works.
--- a/packages/backend/src/admin/admin.module.ts
+++ b/packages/backend/src/admin/admin.module.ts
@@ -2,21 +2,56 @@ import { Module } from '@nestjs/common';
 import { TypeOrmModule } from '@nestjs/typeorm';
 import { ApiKey } from '../entities/api-key.entity';
 import { Tenant } from '../entities/tenant.entity';
+import { Agent } from '../entities/agent.entity';
+import { AgentApiKey } from '../entities/agent-api-key.entity';
+import { AgentEnabledProvider } from '../entities/agent-enabled-provider.entity';
 import { AdminController } from './controllers/admin.controller';
+import { AdminAgentController } from './controllers/admin-agent.controller';
+import { AdminProviderController } from './controllers/admin-provider.controller';
+import { AdminObservabilityController } from './controllers/admin-observability.controller';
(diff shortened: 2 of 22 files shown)


guillaumegay13:
hey @sponsorrockit1, thanks for the PR - it's very interesting.

Before moving forward, I'd like to understand your usage so we could design the best solution.

Would you mind contacting me at guillaume@manifest.build ?

sponsorrockit1:
> **Note:** This comment was posted by my AI agent (ox-alpha / Hermes), working under my GitHub account, in response to the automated review findings.

Thanks to @cubic-dev-ai for the thorough review — all findings were valid and are addressed in d2e8bfc83. Mapping:

**P1**
- **Plaintext admin key in `api_keys.key`** — fixed: the mint service now stores `key = NULL` and relies solely on `key_hash` (salt-aware `verifyKey`), matching the invariant established by HashApiKeys (1771500000000). The raw key is returned exactly once in the create response and never persisted.
- **Bootstrap chicken-and-egg** — addressed with a scoped relaxation rather than a separate surface: a new `@AdminBootstrap()` route decorator lets an authenticated **owner** key call `POST /api/v1/admin/keys` only. A fresh install can now mint its first `ai_admin` key with its existing owner key; every other `/api/v1/admin` route still rejects owner keys with 403, so the admin surface itself stays exclusively `ai_admin`. Happy to discuss alternatives with @guillaumegay13 (env-provisioned first key vs. CLI command) if you'd prefer a different bootstrap mechanism.
- **Unvalidated `authType`** — `@IsIn(['api_key', 'subscription', 'local'])` added to `AttachProviderKeyDto`.

**P2**
- **Partial-failure create** — provider-enablement failure now triggers compensating delete of the just-created agent (mirroring the dashboard controller), then rethrows.
- **Duplicate slug → 500** — mapped to 409 `ConflictException` on create/rename/duplicate, same as the dashboard.
- **Key endpoints returning 200 `{}` for missing agents** — now throw `NotFoundException`.
- **`GET :agentName/key` raw-key readback** — removed; returns `keyPrefix` only. The raw ingest key is one-time at create/rotate/duplicate. (The dashboard's own GET-key handler can decrypt; this admin twin intentionally does not.)
- **Stale dashboard caches after admin mutations** — admin mutations now invalidate both `agentListCacheKey` entries and emit the agent event via `IngestEventBusService`, mirroring the dashboard controller.
- **Legacy rows invisible post-upgrade** (`has_key:false`, verify always false) — two parts:
  - listing reports `has_key = !!key_hash || !!api_key_encrypted`;
  - `verifyKeyMatches` lazily backfills `key_hash` from decryptable credentials on first verify. We deliberately kept migrations catalog-only (no data decryption inside a migration), so upgrades never block boot on undecryptable rows.
- **sameKey branch skipping `key_hash`** — now assigns it on reattach.
- **`SET LOCAL` in non-transactional revert** — both migrations' `down()` use session-scoped `SET lock_timeout` + `RESET`.
- **Serialized usage scans** — parallelized with `Promise.all`.

**P3**
- Prettier width violation in `api-key.guard.ts` fixed.

Verification: backend build green; full jest run **8356 passing**, including updated guard specs covering the new bootstrap path. Three suites (`autofix.module`, `app.config`, `tenant-providers.controller`) fail identically without this commit — env-dependent (`DATABASE_URL`) pre-existing issues.

PR #2762, opened 2026-08-24
fix: refresh harness request counts

### LINKED ISSUE #2758: The harnesses messages' cache is not busted

guillaumegay13:
### ✨ What changed
- Clear both cached harness-list variants before publishing message events.
- Add regression coverage for invalidation ordering.
- Add a patch changeset.

### 💭 Why
The dashboard refreshed after a message event, but GET /agents could still serve cached message counts for up to 60 seconds.

### 👤 For users
Harness request counts now refresh with the next dashboard analytics update.

### 📝 Notes
Closes #2758

---
## Summary by cubic
Refreshes harness request counts by invalidating agent list caches before publishing message events. Previously, `/api/v1/agents` could return cached message counts for up to 60 seconds after a dashboard refresh.

- Invalidate both agent list cache variants (`playground=false|true`) before publishing 'message' events; publishing now awaits deletion to prevent stale counts.
- Adds regression test for invalidation ordering and debounce. No API or config changes; optional cache manager is used when present. Adds a patch changeset for `manifest`. Closes #2758.

<sup>Written for commit 394aa4e8d84ce4c6ae639477416dbaae59f451d4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2762?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/fresh-harness-message-counts.md
+++ b/.changeset/fresh-harness-message-counts.md
@@ -1,0 +1,5 @@
+---
+'manifest': patch
+---
+
+Refresh cached harness message counts when new requests arrive.
--- a/packages/backend/src/analytics/controllers/agents.controller.spec.ts
+++ b/packages/backend/src/analytics/controllers/agents.controller.spec.ts
@@ -10,6 +10,7 @@ import { AgentLifecycleService } from '../services/agent-lifecycle.service';
 import { AgentDuplicationService } from '../services/agent-duplication.service';
 import { ApiKeyGeneratorService } from '../../otlp/services/api-key.service';
 import { TenantCacheService } from '../../common/services/tenant-cache.service';
+import { AgentListCacheService } from '../../common/services/agent-list-cache.service';
 import { IngestEventBusService } from '../../common/services/ingest-event-bus.service';
 import { ProviderService } from '../../routing/routing-core/provider.service';
 import { AutofixStatsService } from '../services/autofix-stats.service';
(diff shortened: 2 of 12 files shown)


guillaumegay13:
Reworked the invalidation after the second cubic pass — the in-flight coordination from `fba5536` is reverted.

**Why the direction changed:** `@UseInterceptors(AgentListCacheInterceptor)` does not resolve the CommonModule singleton. Nest registers an enhancer class in the controller's module `injectables` and instantiates it there (`InterceptorsContextCreator.getInstanceByMetatype` reads `moduleRef.injectables`), so the interceptor serving `GET /agents` was a *different object* from the one injected into the event bus — the in-flight response/write maps could never have lined up. Coordinating on them was both racy (the three findings) and inert.

**What it does now:** the cache key carries a per-tenant generation, and invalidation bumps it (`AgentListCacheService`, a real shared provider). There is no window left to coordinate:

- a request already in flight resolved its key under the old generation, so whenever its response is written back it lands on a key no later reader looks at;
- a request that starts after the bump computes the new key and misses;
- deleting the retired keys is housekeeping, so a cache failure is logged, never thrown — and the message event is **always** published, since a dashboard that never refetches is worse than a count that lags (third finding).

A retired generation is remembered for 2× the 60s list TTL, which is the margin that makes forgetting it safe: nothing writes a retired generation after the bump, so one TTL later those entries have expired.

`AgentsController` now retires keys through the same service instead of deleting them itself. Backend suite green (8440 tests), 100% line and branch coverage on every touched file.

PR #2773, opened 2026-08-27
fix(routing/proxy): emit finish_reason: tool_calls when tool call pre…

crashf:
…cedes empty completion

GPT/Codex streaming can emit `response.output_item.added` with a `function_call` before `response.completed`, whose `output` array is empty. Previously the adapter only checked `output` for function_call items, producing `finish_reason: stop` and causing the client to end the turn prematurely.

- Track `sawToolCall` in `ReasoningStreamState`
- Set `sawToolCall = true` on `response.output_item.added` function_call
- In `handleCompletedEvent`, use `hasFunctionCalls || sawToolCall` to decide `finish_reason`
- Add regression tests for empty-completion + tool-call, text-only completion, and normal function-call completion

Fixes the "GPT no-respond / stops after one tool call" symptom in streaming mode.

---
## Summary by cubic
Fixes the ChatGPT/Codex streaming adapter emitting `finish_reason: stop` when a tool call precedes an empty `response.completed` output, which made clients end the turn before running the tool. The adapter now emits `finish_reason: tool_calls` when any `function_call` was streamed, even if the completed event's output is empty.

- Tracks tool-call presence in the stream transformer state on `response.output_item.added` events.
- Uses the tracked state, alongside existing output inspection, to decide `finish_reason`.
- Adds regression tests for empty-completion tool calls, text-only completions, and normal function-call completions.

<sup>Written for commit f2a798f8e0ec46407b3bc1db4d74894fc23db10f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2773?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/backend/src/routing/proxy/chatgpt-adapter.streaming.spec.ts
+++ b/packages/backend/src/routing/proxy/chatgpt-adapter.streaming.spec.ts
@@ -277,12 +277,11 @@ describe('chatgpt-adapter streaming: finish_reason with empty response.completed
   });
 
   it('keeps finish_reason tool_calls when completed payload still contains the function call', () => {
-    // Pre-existing behavior (completed WITH output listing the function_call)
-    // must not regress.
+    // Isolate the completed-payload path: without a preceding streamed tool
+    // call, this assertion specifically guards hasFunctionCalls.
     const stream = createChatGptStreamTransformer('gpt-5');
     const seen: string[] = [];
     for (const chunk of [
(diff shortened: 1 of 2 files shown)


guillaumegay13:
thanks @crashf!

PR #2786, opened 2026-08-30
feat: configure proxy concurrency limit

daragao3:
### ✨ What changed
- Add `MANIFEST_CONCURRENCY_MAX` with a default of 10.
- Validate plain positive integer values before applying the limit.
- Document the setting for local and Compose deployments.

### 💭 Why
Deployments need to tune per-agent concurrency without source changes.

### 🔧 For operators
Set `MANIFEST_CONCURRENCY_MAX` to a plain positive integer. Invalid values use 10.

### 📝 Notes
The unrelated subscription context change was removed.


guillaumegay13:
Superseded by #2794, which rebuilds the concurrency-only change from current main as one focused commit. Thanks @daragao3 for the original proposal.

PR #2809, opened 2026-09-01
Keep the sidebar bottom stack visible with many harnesses

SebConejo:
✨ What changed
- The harness list is now capped at five and a half rows and scrolls in place
- Scroll shadows appear only on the side where more harnesses remain
- The announcement card and the usage block stay visible on any reasonable window height
- On very short viewports the column scrolls around the still-capped list, so the bottom stack is never far; the mobile drawer keeps its single-column scroll

💭 Why
With many harnesses the list pushed the announcement card below the fold, making its button unreachable without scrolling the whole column.

👤 For users
The bottom of the sidebar stays visible; long harness lists scroll in place with a visual hint.

🔧 For operators
No operator impact.

🧪 How to test
1. Create a dozen harnesses: the list caps at about five and a half rows and scrolls in place
2. The announcement card and usage block stay visible at the bottom
3. On a very short window the column scrolls too, with the list still compact
4. On mobile width, the drawer scrolls as one column as before

### DIFF
--- a/design-system/registry/components.json
+++ b/design-system/registry/components.json
@@ -4503,81 +4503,74 @@
       "line": 784
     },
     {
-      "selector": ".sidebar",
-      "prop": "overflow-y",
-      "value": "auto",
-      "file": "theme.css",
-      "line": 788
-    },
-    {
       "selector": ".sidebar__agent-item",
       "prop": "gap",
       "value": "8px",
       "file": "theme.css",
-      "line": 958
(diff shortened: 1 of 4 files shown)


SebConejo:
Closing without merging: #2811 removes the harness list from the sidebar entirely, so this scroll cap no longer has anything to fix. I rebased the stack so #2810 now targets main directly and no longer carries these commits.

PR #2810, opened 2026-09-01
Record the real origin of waiting-list claims

SebConejo:
✨ What changed
- Waiting-list claims now carry their real origin: cloud or self-hosted, depending on the dashboard that posted them
- A repeat claim with the same email updates the attribution to the latest origin, matching the historical phone-home behavior
- Any other source value is rejected; the website source stays exclusive to the site form

💭 Why
Every claim landed with a generic or stale label, so people joining from the cloud showed up as self-hosted users: wrong data for outreach segmentation.

👤 For users
No visible change; joining the waiting list works exactly as before.

🔧 For operators
Existing mislabeled rows self-correct the next time the person joins again from the right place.

🧪 How to test
1. Join from a cloud dashboard: the claim row reads source cloud
2. Join from a self-hosted dashboard: the row reads self-hosted
3. Join again from the other side: the same row flips to the latest origin

Stacked on #2809; it retargets to main once that one merges.

---
## Summary by cubic
Records the real origin of waiting-list claims so cloud joins no longer show up as self-hosted users.

- Claims carry `cloud` or `self-hosted` matching the dashboard that posted them; a sourceless claim is labeled `cloud` when same-origin, otherwise it defaults to `self-hosted`, and an explicit `null` is rejected.
- A repeat claim with the same email overwrites the previous source, except `website` rows from the site form, which stay untouched.
- The backend rejects any other source value.
- The frontend retries once without the source field if an older backend returns 400, keeping claims working through rollbacks.
- No visible change for users joining the waiting list.

<sup>Written for commit 98900d21bea8511e9ff62c9b4c185d3bd519963e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mnfst/manifest/pull/2810?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/waitlist-claim-source.md
+++ b/.changeset/waitlist-claim-source.md
@@ -1,0 +1,5 @@
+---
+'manifest': patch
+---
+
+Waiting-list claims now record where they were made (cloud or self-hosted) instead of a generic label, and a repeat claim updates the attribution to the latest origin.
--- a/packages/backend/src/waitlist/dto/pivot-claim.dto.spec.ts
+++ b/packages/backend/src/waitlist/dto/pivot-claim.dto.spec.ts
@@ -28,4 +28,11 @@ describe('PivotClaimDto', () => {
     const { errors } = validate({ email: 'jane@example.com', newsletter: true });
     expect(errors.map((error) => error.property)).toContain('newsletter');
   });
+
+  it('accepts the two claim sources and rejects anything else', () => {
+    expect(validate({ email: 'a@b.co', source: 'cloud' }).errors).toHaveLength(0);
+    expect(validate({ email: 'a@b.co', source: 'self-hosted' }).errors).toHaveLength(0);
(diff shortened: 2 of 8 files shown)


SebConejo:
Both cubic findings are handled on the current head. An explicit null source is now rejected with a 400: the default only applies when the field is omitted entirely. On the overwrite concern: rows created by the website form are never touched by this route anymore (the upsert carries a condition excluding source = website). Between cloud and self-hosted the latest claim wins on purpose: both values come from the same person re-joining from a different deployment, and the newest signal is the one we want for outreach.

PR #2879, opened 2026-09-13
feat: add a remote MCP server with OAuth 2.1

guillaumegay13:
Stacked on the **mnfst management CLI** PR #2673. Until that merges, this PR's diff also shows its commits; after it merges, this shrinks to the MCP change. It replaces the fork-only PR (guillaumegay13/manifest#24), which could not run Actions because the fork's account is billing-locked.

### ✨ What changed

- Add a remote MCP server at `POST /api/v1/mcp` over OAuth 2.1.
- Better Auth `jwt`/`mcp`/`cimd` plugins: authorization server, CIMD client identity, resource-bound JWT tokens.
- `requireMcpAuth` gate, RFC 8414/9728 discovery, and the `WWW-Authenticate` challenge.
- Tools reuse the REST/CLI services: harnesses, providers (incl. custom), routing, Autofix/recording, models, pricing, requests, diagnostics.
- Read tools need `mcp:read`; write tools are hidden without `mcp:write`.
- Add the `/consent` page.

### 💭 Why

MCP clients should be able to manage a Manifest workspace, not just route tokens through it.

### 🔧 For operators

- Upgrade `better-auth` 1.6.25 to 1.7.4 (backend and frontend), adding `@better-auth/mcp`, `@better-auth/cimd`, `@better-auth/oauth-provider`, `@modelcontextprotocol/server`.
- OAuth/JWKS tables are created by the existing boot-time Better Auth migration.
- `BETTER_AUTH_URL` determines the issuer and MCP resource for self-hosted installs.

### 📝 Notes

- Docs: mnfst/docs#99.


benbarlev10-stack:
The `mcp:read`/`mcp:write` split is a good coarse gate, but both are decided at token issuance, not at call time, and that gap is worth closing before the CIMD path hardens. With `mcp:write`, `tools/list` advertises every write tool to any client holding the scope; nothing at the MCP boundary then binds a specific `tools/call` to the principal, the resource (which workspace), and the exact args. Two cases to check:

1. `tools/list` vs `tools/call`: the OAuth resource claim can name one workspace while the tool's data model lets it operate on another. A coarse scope is a ceiling a call can exceed.
2. A denied call should be a written decision, not an absence. Emit the principal, tool, args-hash, and the OAuth resource claim on the same record at the enforcement point, so a rejection is reconstructable and the caller cannot rewrite it.

I build the same boundary in AgentKey (authorization + evidence per MCP/agent call, https://agentkey.us); the per-call decision record tying scope to tool+args+resource is exactly the piece this PR is positioned to add.

